PR #1, opened 2016-08-25
Add method and URL to Event, add method to EventDispatcher

delikat:
- Adds method and URL properties to `Event`
- Adds method argument to `EventDispatcher.dispatch_event`
- Renames `EventBuilder` to `EventBuilderV1`
- Removes abstract `BaseEventBuilder` class (https://github.com/bbatsov/ruby-style-guide#duck-typing)
- Locks down a few unnecessary `attr_accessors` to `attr_readers`
- Fixes broken tests

@aliabbasrizvi @delikat @haleybash-optimizely @mikeng13 @alda-optimizely @wangjoshuah @jophde @jsm

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -70,7 +70,7 @@ module Optimizely
       @logger.log(Logger::INFO,
                   'Dispatching impression event to URL %s with params %s.' % [impression_event.url,
                                                                               impression_event.params])
-      @event_dispatcher.dispatch_event(impression_event.method, impression_event.url, impression_event.params)
+      @event_dispatcher.dispatch_event(impression_event.url, impression_event.params, impression_event.http_verb)
 
       @config.get_variation_key_from_id(experiment_key, variation_id)
     end
@@ -136,7 +136,7 @@ module Optimizely
       @logger.log(Logger::INFO,
                   'Dispatching conversion event to URL %s with params %s.' % [conversion_event.url,
                                                                               conversion_event.params])
-      @event_dispatcher.dispatch_event(conversion_event.method, conversion_event.url, conversion_event.params)
+      @event_dispatcher.dispatch_event(conversion_event.url, conversion_event.params, conversion_event.http_verb)
     end
 
     private
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -70,7 +70,7 @@ module Optimizely
       @logger.log(Logger::INFO,
                   'Dispatching impression event to URL %s with params %s.' % [impression_event.url,
                                                                               impression_event.params])
-      @event_dispatcher.dispatch_event(impression_event.url, impression_event.params, impression_event.http_verb)
+      @event_dispatcher.dispatch_event(impression_event)
 
       @config.get_variation_key_from_id(experiment_key, variation_id)
     end
@@ -136,7 +136,7 @@ module Optimizely
       @logger.log(Logger::INFO,
                   'Dispatching conversion event to URL %s with params %s.' % [conversion_event.url,
                                                                               conversion_event.params])
-      @event_dispatcher.dispatch_event(conversion_event.url, conversion_event.params, conversion_event.http_verb)
+      @event_dispatcher.dispatch_event(conversion_event)
     end
 
     private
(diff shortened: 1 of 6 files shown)


lib/optimizely/event_builder.rb
```diff
@@ -6,28 +6,18 @@ module Optimizely
   class Event
     # Representation of an event which can be sent to the Optimizely logging endpoint.
 
-    # Event API format
-    OFFLINE_API_PATH = 'https://%{project_id}.log.optimizely.com/event'
-
-    # Gets/Sets event params.
-    attr_accessor :params
+    attr_reader :method
```
mikeng13: these all need docs

lib/optimizely/event_dispatcher.rb
```diff
@@ -1,33 +1,30 @@
...
+    def dispatch_event(_method, _url, _params)
     end
   end
 
-  class EventDispatcher < BaseEventDispatcher
+  class EventDispatcher
     REQUEST_TIMEOUT = 10
 
-    def dispatch_event(url, params)
+    def dispatch_event(method, url, params)
```
aliabbasrizvi: Lets make this the third argument and optional

lib/optimizely/event_builder.rb
delikat: Good call, I like `http_verb`.

lib/optimizely.rb
```diff
@@ -70,7 +70,7 @@ def activate(experiment_key, user_id, attributes = nil)
       @logger.log(Logger::INFO,
                   'Dispatching impression event to URL %s with params %s.' % [impression_event.url,
                                                                               impression_event.params])
-      @event_dispatcher.dispatch_event(impression_event.url, impression_event.params)
+      @event_dispatcher.dispatch_event(impression_event.method, impression_event.url, impression_event.params)
```
mikeng13: this is more of a design thing, but semantically speaking, when you say dispatch_event, it makes more sense to pass in the event object instead of splitting it up. As we evolve the SDK I can see this method signature becoming too long/unbearable. Might as well make the refactor here.

lib/optimizely/event_builder.rb
```diff
@@ -6,28 +6,18 @@ module Optimizely
...
-      # project_id - ID for the project.
-      #
-      # Returns URL for event API.
-
-      sprintf(OFFLINE_API_PATH, project_id: @params[Params::PROJECT_ID])
-    end
   end
 
-  class EventBuilder
+  class EventBuilderV1
```
mikeng13: Instead of creating two different event builder classes, why don't we have one event builder that can build V1 events and V2 events. That way you don't have to worry about handling the instantiation of two different builder classes.

lib/optimizely.rb
delikat: Agreed, was going to refactor to pass around `Event`s in a later PR but will do so here.

lib/optimizely/event_builder.rb
delikat: Because when we eventually remove V1 support it will be easier to simply remove the `EventBuilderV1` class than to rip out the V1 logic from a larger omnibus `EventBuilder`.

spec/event_builder_spec.rb
```diff
@@ -52,7 +48,7 @@
 
     expect(@event_builder).to receive(:create_impression_event)
                           .with('test_experiment', 'test_user')
-                          .and_return(Optimizely::Event.new(expected_params))
+                          .and_return(Optimizely::Event.new(:get, '', expected_params))
```
mikeng13: for completeness, we should also have a test case to check that the event is built with the correct method and url

lib/optimizely/event_builder.rb
mikeng13: I disagree because when we rip out V1 support, we would just be changing the internals of this class and the API references. If we keep two classes around there will also be some forking logic throughout the codebase that we need to kill. Providing the event builder as  some sort of facade/factory will make it much easier to deal with.

aliabbasrizvi: I think it is way better to pick the right EventBuilder at object invocation and not have logic all over. The model here allows that.

mikeng13: How about this, since the `EventBuilder` is already initialized with the project config (which tells us whether we are in V1 or V2), why don't we just have the event builder determine whether to return a V1 Event or V2 event when we call eventBuilder.getConversionEvent and eventBuilder.getImpressionEvent. That event can be passed into event dispatcher, which determines based on the event whether to get or post

aliabbasrizvi:
LGTM

PR #25, opened 2017-02-03
Allow httparty 0.14

gaganawhad:
This updates the gemspec to allow httparty 0.14


gaganawhad:
@delikat Thanks! 🎉 .  The agreement should be signed now.

PR #29, opened 2017-02-09
Relax patch version requirements on runtime dependencies

gaganawhad:
It generally helps to be optimistic about patch version requirements on
runtime dependencies used in a gem. This should help reduce version conflicts when
using this gem.


gaganawhad:
Allright @delikat ... you allowed me to contribute, so that might have made me a bit trigger happy here :) Just wanted to submit this one as well and see what you think. This is in the same spirit of what we did with HTTParty. I was getting some version conflicts with `json-schema`.

PR #68, opened 2017-10-09
Feature - Bucketing IDs

oakbani:
🖊  Please note that since data file in Ruby SDK is different from PHP-SDK, the resulting experiment/variation values and the flow of resulting log statements due to bucketing may differ in the unit tests.

### DIFF
--- a/spec/bucketing_spec.rb
+++ b/spec/bucketing_spec.rb
@@ -147,14 +147,15 @@ describe Optimizely::Bucketer do
 
   describe 'Bucketing with Bucketing Id' do
     # Bucketing with bucketing ID
+    # Make sure that the bucketing ID is used for the bucketing and not the user ID
     it 'should bucket to a variation different than the one expected with the userId' do
       experiment = config.get_experiment_from_key('test_experiment')
 
-      # Bucketing with user id as bucketing id - 'test_user111127' produces bucket value < 5000
+      # Bucketing with user id as bucketing id - 'test_user111127' produces bucket value < 5000 thus buckets to control
       expected_variation = config.get_variation_from_id('test_experiment','111128')
       expect(bucketer.bucket(experiment,'test_user', 'test_user')).to be(expected_variation)
 
-      # Bucketing with bucketing id - 'any_string789111127' produces bucket value btw 5000 to 10,000
+      # Bucketing with bucketing id - 'any_string789111127' produces bucket value btw 5000 to 10,000 thus buckets to variation
       expected_variation = config.get_variation_from_id('test_experiment','111129')
       expect(bucketer.bucket(experiment,'any_string789', 'test_user')).to be(expected_variation)
     end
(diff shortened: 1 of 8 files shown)


spec/bucketing_spec.rb
```diff
@@ -108,47 +108,82 @@ def get_bucketing_id(user_id, entity_id=nil)
...
+      experiment = config.get_experiment_from_key('group1_exp1')
+
+      expected_variation = config.get_variation_from_id('group1_exp1','130001')
+      expect(bucketer.bucket(experiment,'pid', 'test_user')).to be(expected_variation)
+
+      expected_variation = config.get_variation_from_id('group1_exp1','130002')
+      expect(bucketer.bucket(experiment,'123456789', 'test_user')).to be(expected_variation)
+    end
+  end
+
```
alda-optimizely: Please add the cases when the bucketing key fallbacks to the user ID. This happens when the bucketing key is null. So, to clarify:
• Bucketing with bucketing ID
-Non null bucketing ID
-Null bucketing ID
• Bucketing with invalid experiment key and bucketing ID
-Non null bucketing ID
-Null bucketing ID
• Bucketing with grouped experiments and bucketing ID
-Non null bucketing ID
-Null bucketing ID

I've updated the doc with these corner cases. 

You can look at the PHP example: 
https://github.com/optimizely/php-sdk/blob/master/tests/BucketerTest.php#L298

spec/bucketing_spec.rb
```diff
@@ -75,9 +75,9 @@ def get_bucketing_id(user_id, entity_id=nil)
     expect(bucketer).to receive(:generate_bucket_value).once.and_return(3000)
 
     experiment = config.get_experiment_from_key('group1_exp2')
-    expect(bucketer.bucket(experiment, 'test_user')).to be_nil
+    expect(bucketer.bucket(experiment,'bucket_id_ignored', 'test_user')).to be_nil
```
alda-optimizely: what does the bucketing key: `bucketing_id_ignored` do?
how is it ignored?

spec/decision_service_spec.rb
```diff
@@ -162,6 +195,31 @@
                           .with(Logger::INFO, "Saved variation ID 111128 of experiment ID 111127 for user 'test_user'.")
       end
 
+       it 'should look up the UserProfile, bucket normally (using Bucketing ID attrbiute), and save the result if no saved profile is found' do
```
alda-optimizely: sp: attribute

spec/bucketing_spec.rb
oakbani: I have used this string in all those cases where we mock  the method generate_bucket_value to test traffic allocations as in above
expect(bucketer).to receive(:generate_bucket_value).twice.and_return(3000)
This makes it more readable to the developer that the string passed as bucketing id does not have an impact

spec/bucketing_spec.rb
oakbani: @alda-optimizely  I have a question regarding this. Please see my comments here:
https://github.com/msohailhussain/php-sdk/pull/6

alda-optimizely: Yea you're right. I don't remember why I did that...maybe it was some legacy implementation that I forgot to update? The main goal for the test is to make sure that the bucketing ID is being used and not the userID for the bucketing. So, userID should bucket to the non expected variation. I've updated my PHP tests.

oakbani:
@alda-optimizely  Updated with required changes. Let me know if there is anything else

PR #73, opened 2017-11-09
Update Ruby SDK to return nil when incorrect API is used to fetch a variable

rashidsp:


wangjoshuah:
build

lib/optimizely/decision_service.rb
```diff
@@ -156,65 +145,43 @@ def get_variation_for_feature_experiment(feature_flag, user_id, attributes = nil
...
-          group = @config.group_key_map[group_id]
-          bucketed_experiment_id = @bucketer.find_bucket(user_id, group_id, group['trafficAllocation'])
-          if bucketed_experiment_id.nil?
+      if feature_flag['experimentIds'].empty?
+        @config.logger.log(
+          Logger::DEBUG,
+          "The feature flag '#{feature_flag_key}' is not used in any experiments."
+        )
+        return nil
+      else
```
mikeng13: nitpick: because you return out of the previous `if` clause, you can skip the `else` so we reduce the nesting

PR #81, opened 2017-12-08
1.5.0 bump

alda-optimizely:


alda-optimizely:
The gemspec refers to `VERSION` which is defined in https://github.com/optimizely/ruby-sdk/pull/81/files#diff-4ee21fe22c6530b8ca5cc6cbba691aca.

PR #87, opened 2018-02-14
Feature toggleability implementation

rashidsp:


lib/optimizely.rb
```diff
@@ -263,23 +263,26 @@ def is_feature_enabled(feature_flag_key, user_id, attributes = nil)
...
-          @logger.log(Logger::DEBUG,
-                      "The user '#{user_id}' is not being experimented on in feature '#{feature_flag_key}'.")
+      if decision.nil?
+        @logger.log(Logger::INFO,
+                    "Feature '#{feature_flag_key}' is not enabled for user '#{user_id}'.")
+        return false
+      end
+      variation = decision['variation']
+      if decision.source == Optimizely::DecisionService::DECISION_SOURCE_EXPERIMENT
+        unless variation['featureEnabled']
```
mikeng13: This check shouldn't be here. We also use `featureEnabled` for rollout variations.

spec/project_spec.rb
```diff
@@ -676,6 +676,22 @@ class InvalidErrorHandler; end
       expect(spy_logger).to have_received(:log).once.with(Logger::INFO, "Feature 'boolean_single_variable_feature' is enabled for user 'test_user'.")
     end
 
+    it 'should return true if the user is bucketed into a feature rollout even if featureEnabled property is false' do
```
mikeng13: This is not the case. We should always be checking the `featureEnabled` property. The reason is because in the future we might conceive of a world where there "could" be multiple variations. In the datafile the property will ALWAYS be set to true if there is traffic allocated to it.

oakbani:
@mikeng13  Your comments have been addressed here https://github.com/optimizely/ruby-sdk/pull/88. Closing this one

PR #90, opened 2018-02-27
Parse numeric float/string as integer and add logger

oakbani:
* Revenue value parsed as in https://github.com/optimizely/php-sdk/pull/91
* Added logger to EventBuilder


mikeng13:
build

PR #103, opened 2018-04-19
Implements inputs validator

rashidsp:
Implements Null and String validator for inputs.
inputs : ('feature_flag_key', 'experiment_key', 'user_id', 'variation_key', 'variable_key' and 'variable_type')
Methods:
-> Activate 
-> get_variation
-> Track
-> set_forced_variation through project_config-> set_forced_variation
-> get_forced_variation through project_config-> get_forced_variation
-> is_feature_enabled
-> get_enabled_features
-> get_feature_variable_for_type

### DIFF
--- a/lib/optimizely/helpers/validator.rb
+++ b/lib/optimizely/helpers/validator.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 #
-#    Copyright 2016-2017, Optimizely and contributors
+#    Copyright 2016-2018, Optimizely and contributors
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


mikeng13:
Will be looking at this soon. Wanted to see if maybe we could run this through the compat suite

lib/optimizely/helpers/validator.rb
```diff
@@ -95,6 +95,27 @@ def string_numeric?(str)
...
+        #
+        # logger - logger.
+        #
+        # Returns boolean True if all of the values are valid, False otherwise.
+
+        return false unless variables.respond_to?(:each) && !variables.empty?
+        is_valid = true
+        variables.each do |key, value|
+          next if value.is_a?(String) && !value.empty?
+          is_valid = false
```
thomaszurkan-optimizely: I don't think this is correct and your tests seem to confirm this.  If the last argument is invalid, everything is fine,  but, if your middle argument is invalid, it would not catch it.  Shouldn't it be is_valid ||= false?  Or return when you get the first is_valid = false.  Also, can you add tests that actually call the activate with a integer user id or a empty experiment key (first argument)?

thomaszurkan-optimizely: please put a break after setting is valid to false.

PR #109, opened 2018-06-14
Add datafile isValid check in feature accessor methods

### LINKED ISSUE #107: NoMethodError: undefined method `get_feature_flag_from_key` for nil:NilClass when calling `get_feature_variable_*`
SDK version 2.0.1

On occasions where fetching a valid datafile fails, any call to a `get_feature_variable_*` method will raise the exception `NoMethodError: undefined method 'get_feature_flag_from_key' for nil:NilClass` instead of failing gracefully and logging.

Within `Project#get_feature_variable_for_type` there is no check for if the project is useable via the [`is_valid` attr_reader](https://github.com/optimizely/ruby-sdk/blob/c1ca00d7a73581fd5fbafff692e343490d8448b3/lib/optimizely.rb#L35) before attempting to access the config

https://github.com/optimizely/ruby-sdk/blob/c1ca00d7a73581fd5fbafff692e343490d8448b3/lib/optimizely.rb#L439-L449

All of the other methods seem to return nil or false if the datafile cannot be parsed

https://github.com/optimizely/ruby-sdk/blob/c1ca00d7a73581fd5fbafff692e343490d8448b3/lib/optimizely.rb#L262-L266

oakbani:
Fixes #107

### DIFF
--- a/spec/project_spec.rb
+++ b/spec/project_spec.rb
@@ -893,6 +893,17 @@ describe 'Optimizely' do
     user_id = 'test_user'
     user_attributes = {}
 
+    it 'should return nil when called with invalid project config' do
+      logger = double('logger')
+      allow(logger).to receive(:log)
+      allow(Optimizely::SimpleLogger).to receive(:new) { logger }
+      invalid_project = Optimizely::Project.new('invalid', nil, spy_logger)
+      expect(invalid_project.get_feature_variable_string('string_single_variable_feature', 'string_variable', user_id, user_attributes))
+        .to eq(nil)
+      expect(logger).to have_received(:log).once.with(Logger::ERROR, 'Provided datafile is in an invalid format.')
+      expect(logger).to have_received(:log).once.with(Logger::ERROR, 'Provided datafile is in an invalid format. Aborting get_feature_variable_string.')
+    end
+
     describe 'when the feature flag is enabled for the user' do
       describe 'and a variable usage instance is not found' do
         it 'should return the default variable value' do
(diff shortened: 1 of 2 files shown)


spec/project_spec.rb
```diff
@@ -1029,6 +1040,17 @@ class InvalidErrorHandler; end
...
     user_attributes = {}
 
+    it 'should return nil when called with invalid project config' do
+      logger = double('logger')
+      allow(logger).to receive(:log)
+      allow(Optimizely::SimpleLogger).to receive(:new) { logger }
+      invalid_project = Optimizely::Project.new('invalid', nil, spy_logger)
+      expect(invalid_project.get_feature_variable_boolean('boolean_single_variable_feature', 'boolean_variable', user_id, user_attributes))
+        .to eq(nil)
+      expect(logger).to have_received(:log).once.with(Logger::ERROR, 'Provided datafile is in an invalid format.')
```
mikeng13: Is this right? How come we use similar messages twice?

spec/project_spec.rb
```diff
@@ -1055,6 +1077,17 @@ class InvalidErrorHandler; end
...
     user_attributes = {}
 
+    it 'should return nil when called with invalid project config' do
+      logger = double('logger')
+      allow(logger).to receive(:log)
+      allow(Optimizely::SimpleLogger).to receive(:new) { logger }
+      invalid_project = Optimizely::Project.new('invalid', nil, spy_logger)
+      expect(invalid_project.get_feature_variable_double('double_single_variable_feature', 'double_variable', user_id, user_attributes))
+        .to eq(nil)
+      expect(logger).to have_received(:log).once.with(Logger::ERROR, 'Provided datafile is in an invalid format.')
```
mikeng13: Shouldn't we only be logging the second message?

spec/project_spec.rb
```diff
@@ -1082,6 +1115,17 @@ class InvalidErrorHandler; end
...
 
+    it 'should return nil when called with invalid project config' do
+      logger = double('logger')
+      allow(logger).to receive(:log)
+      allow(Optimizely::SimpleLogger).to receive(:new) { logger }
+      invalid_project = Optimizely::Project.new('invalid', nil, spy_logger)
+      expect(invalid_project.get_feature_variable_integer('integer_single_variable_feature', 'integer_variable', user_id, user_attributes))
+        .to eq(nil)
+      expect(logger).to have_received(:log).once.with(Logger::ERROR, 'Provided datafile is in an invalid format.')
+      expect(logger).to have_received(:log).once.with(Logger::ERROR, 'Provided datafile is in an invalid format. Aborting get_feature_variable_integer.')
```
mikeng13: Same as above, seems redundant

spec/project_spec.rb
oakbani: The first one comes from the initialize method. And the second one is logged in the relevant API method. Though duplicate, I think we should keep this at both places for clarity.
If a client is initialized, it should log the datafile error. When an API method is called, it should log exactly why the method is aborting. 
Let me know if we should revise the error log in the API method. And shorten it to only 'Aborting <method name>"

mikeng13: Yeah that makes sense. I agree with your assessment.

PR #113, opened 2018-06-25
Rashid/docs

rashidsp:
The generated docs [gh-pages branch](https://github.com/optimizely/ruby-sdk/tree/gh-pages)


rashidsp:
@mikeng13 In that case modules will have to be set to public, which will eventually include all Inner methods and attributes, then we would need to exclude them explicitly.

PR #116, opened 2018-06-26
Use parens to prevent string literal in condition

skipsuva:
We just integrated this gem into our Rails project at work, and have begun seeing the following warning when booting the app (via docker) or running the suite:

`/usr/local/bundle/gems/optimizely-sdk-2.0.1/lib/optimizely/project_config.rb:90: warning: string literal in condition`

Our suspicion is that the ruby interpreter wasn't properly parsing the lack of parens on the `.key?` method call in: `@anonymize_ip = config.key? 'anonymizeIP' ? config['anonymizeIP'] : false` 

Lemme know what you think.  As I said, we just started using this gem but it's been working great so far - thanks for the work!

### DIFF
--- a/lib/optimizely/project_config.rb
+++ b/lib/optimizely/project_config.rb
@@ -87,7 +87,7 @@ module Optimizely
       @feature_flags = config.fetch('featureFlags', [])
       @groups = config.fetch('groups', [])
       @project_id = config['projectId']
-      @anonymize_ip = config.key? 'anonymizeIP' ? config['anonymizeIP'] : false
+      @anonymize_ip = config.key?('anonymizeIP') ? config['anonymizeIP'] : false
       @revision = config['revision']
       @rollouts = config.fetch('rollouts', [])
 


PatrickTulskie:
I just tested this theory as well since I'm also seeing this in development...

    config = { 'anonymizeIP' => true }

    def test?(string)
      true
    end

    puts test? 'anonymizeIP' ? config['anonymizeIP'] : false
    puts test?('anonymizeIP') ? config['anonymizeIP'] : false

Yields the following output:

    untitled:7: warning: string literal in condition
    true
    true

This PR will fix the warning.

mikeng13:
Thanks for submitting this fix! Will review and run our tests on it shortly.

mikeng13:
build

skipsuva:
Awesome, thanks!

mikeng13:
Of course, thanks for submitting! 

Btw, even though I merged it in, we won't yet be deploying this. We have an upcoming minor version release 2.1 that will include this change and some new features in the coming weeks.

PR #118, opened 2018-06-27
Update condition.rb

tomfuertes:
Not sure if this is the right place to log this, but I'm working with the Javascript SDK and getting the right result back, however, the Ruby SDK is returning a nil value for someone in the right audience.

I'm sure there is a more elegant fix for this and/or something up the chain might have changed, but right now for the hello world use case on install this is broken for me

```
key = 'NEW_USER'
obj = {:NEW_USER=> 'true'}
p obj[key] # prints nil but should print 'true'
```

### DIFF
--- a/lib/optimizely/condition.rb
+++ b/lib/optimizely/condition.rb
@@ -89,7 +89,7 @@ module Optimizely
       #
       # Returns boolean indicating the result of comparing the condition value against the user attributes.
 
-      condition_array[1] == @user_attributes[condition_array[0]]
+      condition_array[1] == @user_attributes[condition_array[0].to_sym]
     end
 
     def evaluate(conditions)


mikeng13:
@tomfuertes this fails because you are using a symbol as the key. Try this instead: 

```key = "NEW_USER"
obj = { key => 'true'}
```

tomfuertes:
@mikeng13 Yes, I get that. The use case is that if you pass `{KEY: value}` to activate and not `{KEY=> value}` you get a silent fail condition that could save the next client ~25 hours of debugging and a QA/sprint delay to deploy:

Fail: 

image


Success:

image

mikeng13:
Closing as inactive

PR #120, opened 2018-08-27
fix(track): Send decisions for all experiments using an event when using track

rashidsp:


mikeng13:
build

spec/event_builder_spec.rb
```diff
@@ -253,6 +253,18 @@
     expect(conversion_event.http_verb).to eq(:post)
   end
 
+  it 'should create_conversion_event when Event is used in multiple experiments' do
+    @expected_conversion_params[:visitors][0][:snapshots][0][:decisions].push(
```
mikeng13: The setup of this test is not very clear from just looking at this. Can you split this out into its own scenario with its own setup? Specifically, the scenario testing event tracking with multiple experiments.

lib/optimizely/event_builder.rb
```diff
@@ -195,43 +195,38 @@ def get_conversion_params(event_key, event_tags, experiment_variation_map)
...
+      if event_tags
+        revenue_value = Helpers::EventTagUtils.get_revenue_value(event_tags, @logger)
+        event_object[:revenue] = revenue_value if revenue_value
 
-        single_snapshot[:events] = [event_object]
+        numeric_value = Helpers::EventTagUtils.get_numeric_value(event_tags, @logger)
+        event_object[:value] = numeric_value if numeric_value
 
-        conversion_event_params.push(single_snapshot)
+        event_object[:tags] = event_tags
```
msohailhussain: Please add a check if event_tags has any item in it.

PR #124, opened 2018-08-30
fix(datafile-parsing): Prevent newer versions datafile

rashidsp:


lib/optimizely.rb
```diff
@@ -64,17 +64,15 @@ def initialize(datafile, event_dispatcher = nil, logger = nil, error_handler = n
 
       begin
         @config = ProjectConfig.new(datafile, @logger, @error_handler)
-      rescue
+      rescue InvalidDatafileVersionError => e
```
mikeng13: Same comment as on the Python PR, please consolidate into one `rescue` and change the message based on the type of the exception

lib/optimizely.rb
```diff
@@ -64,17 +64,15 @@ def initialize(datafile, event_dispatcher = nil, logger = nil, error_handler = n
...
+        @logger.log(Logger::ERROR, e.message)
         return
-      end
-
-      unless @config.parsing_succeeded?
+      rescue
         @is_valid = false
         @logger = SimpleLogger.new
-        @logger.log(Logger::ERROR, InvalidDatafileVersionError.new.message)
+        @logger.log(Logger::ERROR, InvalidInputError.new('datafile').message)
```
mikeng13: Also, please pass the error to the `error_handler` instance

mikeng13:
Ran these compat-suite tests locally and they pass: https://github.com/optimizely/fullstack-sdk-compatibility-suite/pull/134

PR #125, opened 2018-09-17
feat(api): Accepting all types for attributes values

rashidsp:
**Note:** https://github.com/optimizely/ruby-sdk/blob/0e9980a39e81a19cb368c2188b135d7450551516/lib/optimizely/helpers/validator.rb#L43

ATTRIBUTE_VALID_TYPES.include?(attribute_value.class) can't be used as Fixnum and Bignum are deprecated in ruby 2.4 and Rubocop fails: Lint/UnifiedInteger: Use Integer instead of Fixnum.

### DIFF
--- a/lib/optimizely/decision_service.rb
+++ b/lib/optimizely/decision_service.rb
@@ -362,17 +362,16 @@ module Optimizely
       #
       # user_id - String user ID
       # attributes - Hash user attributes
-      # By default, the bucketing ID should be the user ID
-      bucketing_id = user_id
+      # Returns String representing bucketing ID if it is a String type in attributes else return user ID
 
-      # If the bucketing ID key is defined in attributes, then use that in place of the userID
-      if attributes && attributes[Optimizely::Helpers::Constants::CONTROL_ATTRIBUTES['BUCKETING_ID']].is_a?(String)
-        unless attributes[Optimizely::Helpers::Constants::CONTROL_ATTRIBUTES['BUCKETING_ID']].empty?
-          bucketing_id = attributes[Optimizely::Helpers::Constants::CONTROL_ATTRIBUTES['BUCKETING_ID']]
-          @config.logger.log(Logger::DEBUG, "Setting the bucketing ID '#{bucketing_id}'")
-        end
+      attributes ||= {}
+      bucketing_id = attributes[Optimizely::Helpers::Constants::CONTROL_ATTRIBUTES['BUCKETING_ID']]
+
+      if bucketing_id
+        return bucketing_id if bucketing_id.is_a?(String)
+        @config.logger.log(Logger::WARN, 'Bucketing ID attribute is not a string. Defaulted to user ID.')
       end
-      bucketing_id
+      user_id
     end
   end
 end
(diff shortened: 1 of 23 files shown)


mikeng13:
Going to wait on this until we resolve and merge the Python PR, then we should make sure this matches the requirements over there as well @msohailhussain

msohailhussain:
build

lib/optimizely/event_builder.rb
```diff
@@ -74,9 +76,9 @@ def get_common_params(user_id, attributes)
       visitor_attributes = []
 
       attributes&.keys&.each do |attribute_key|
-        # Omit null attribute values
+        # Omit invalid attribute values.
```
msohailhussain: correct comment is "Omit attribute values that are not supported by the log endpoint."
we need to check attribute key is string or not?

spec/condition_spec.rb
```diff
@@ -145,4 +145,19 @@
     condition = JSON.parse(condition)
     expect(@condition_evaluator.evaluate(condition)).to be true
   end
+
+  it 'should evaluate to true for evaluate when user attributes evaluate true' do
```
mikeng13: Can you re-word this: `should evaluate to true when user attributes match the audience conditions`

spec/event_builder_spec.rb
```diff
@@ -125,6 +125,70 @@
     expect(impression_event.http_verb).to eq(:post)
   end
 
+  it 'should create a valid Event when create_impression_event is called with attributes of different valid types' do
+    @expected_impression_params[:visitors][0][:attributes].unshift(
```
mikeng13: Can you copy this into a new variable and mutate that one for use instead?

spec/event_builder_spec.rb
```diff
@@ -125,6 +125,70 @@
...
+      'double_key' => 5.5
+    }
+    impression_event = @event_builder.create_impression_event(experiment, '111128', 'test_user', attributes)
+    expect(impression_event.params).to eq(@expected_impression_params)
+    expect(impression_event.url).to eq(@expected_endpoint)
+    expect(impression_event.http_verb).to eq(:post)
+  end
+
+  it 'should create a valid Event and exclude attributes of invalid types' do
+    @expected_impression_params[:visitors][0][:attributes].unshift(
```
mikeng13: Same as above

lib/optimizely/decision_service.rb
```diff
@@ -362,17 +362,16 @@ def get_bucketing_id(user_id, attributes)
...
-      bucketing_id = user_id
+      # Returns String representing bucketing ID if it is a String type in attributes else return user ID
 
-      # If the bucketing ID key is defined in attributes, then use that in place of the userID
-      if attributes && attributes[Optimizely::Helpers::Constants::CONTROL_ATTRIBUTES['BUCKETING_ID']].is_a?(String)
-        unless attributes[Optimizely::Helpers::Constants::CONTROL_ATTRIBUTES['BUCKETING_ID']].empty?
-          bucketing_id = attributes[Optimizely::Helpers::Constants::CONTROL_ATTRIBUTES['BUCKETING_ID']]
-          @config.logger.log(Logger::DEBUG, "Setting the bucketing ID '#{bucketing_id}'")
-        end
+      attributes ||= {}
```
mikeng13: Let's return the user_id if attributes is not defined or null

mikeng13:
@rashidsp looks like the test failures are legit. There are some lint errors.

PR #133, opened 2018-10-16
Test against ruby version(s)

apanzerj:
Description:

Some small housekeeping.

1. Test against latest minor versions of ruby (excluding 2.6, which is not published)

2. Unpin stuff. This is a general rule (IMO) in the ruby world. You don't pin things unless you have a reason. 

3. Fix small style violations

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,16 @@
+branch:
+  only:
+    - master
 language: ruby
 rvm:
-  - 2.3.1
-before_install: gem install bundler -v 1.10.6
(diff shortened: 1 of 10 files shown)


apanzerj:
The only failure was SourceClear for some reason.

apanzerj:
aaaaand this is why we don't pin things. I'll look at the style violations and make changes in a bit.

apanzerj:
Hey! When you actually include the coveralls gem it works.

PR #136, opened 2018-11-06
Fix: Lint errors after rubocop update.

rashidsp:
**Summary**

- In PR https://github.com/optimizely/ruby-sdk/pull/133 Rubocop's latest version 0.60.0 was installed, due to which later PRs were getting failed with lint errors.
- Configured rubocop with latest version ` 0.60.0`.
- Fixed lint errors.

### DIFF
--- a/lib/optimizely/bucketer.rb
+++ b/lib/optimizely/bucketer.rb
@@ -92,9 +92,7 @@ module Optimizely
       end
 
       # Handle the case when the traffic range is empty due to sticky bucketing
-      if variation_id == ''
-        @config.logger.log(Logger::DEBUG, 'Bucketed into an empty traffic range. Returning nil.')
-      end
+      @config.logger.log(Logger::DEBUG, 'Bucketed into an empty traffic range. Returning nil.') if variation_id == ''
 
       @config.logger.log(Logger::INFO, "User '#{user_id}' is in no variation.")
       nil
--- a/lib/optimizely/bucketer.rb
+++ b/lib/optimizely/bucketer.rb
@@ -92,7 +92,12 @@ module Optimizely
       end
 
       # Handle the case when the traffic range is empty due to sticky bucketing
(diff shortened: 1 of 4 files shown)


lib/optimizely/helpers/validator.rb
```diff
@@ -112,9 +112,7 @@ def inputs_valid?(variables, logger = NoOpLogger.new, level = Logger::ERROR)
           next if value.is_a?(String) && !value.empty?
 
           is_valid = false
-          if logger_valid?(logger) && level
-            logger.log(level, "#{Optimizely::Helpers::Constants::INPUT_VARIABLES[key.to_s.upcase]} is invalid")
-          end
+          logger.log(level, "#{Optimizely::Helpers::Constants::INPUT_VARIABLES[key.to_s.upcase]} is invalid") if logger_valid?(logger) && level
```
mikeng13: Hmm, I find this is harder to read than before, is that really one of our int rules? If so we should change it so that it won't complain about this. @apanzerj what are your thoughts on this kind of styling rule?

rashidsp: Plz review the addressed change, it's also complying with rubocop rules.

lib/optimizely/bucketer.rb
```diff
@@ -92,9 +92,7 @@ def bucket(experiment, bucketing_id, user_id)
       end
 
       # Handle the case when the traffic range is empty due to sticky bucketing
-      if variation_id == ''
-        @config.logger.log(Logger::DEBUG, 'Bucketed into an empty traffic range. Returning nil.')
-      end
+      @config.logger.log(Logger::DEBUG, 'Bucketed into an empty traffic range. Returning nil.') if variation_id == ''
```
mikeng13: Same comment here, is there a way to invert this relationship? Seems like it makes it harder to read

oakbani:
@rashidsp I would suggest that rather than modifying code or adding extra lines to comply with rubocop rule, ignore this particular rule in rubocop.yml.

PR #138, opened 2018-11-13
refac(ci): Trigger fullstack-compat-test on travis

juancarlostong:
# Summary
- Runs the fullstack compat suite tests on Travis CI
- Accompanying PR: https://github.com/optimizely/fullstack-sdk-compatibility-suite/pull/158

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,3 +14,17 @@ before_script: "rubocop"
 addons:
   srcclr: true
 after_success: "coveralls"
+
+jobs:
+  include:
+    - stage: 'Integration tests'
+      env: SDK=ruby
+      language: python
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,3 +14,17 @@ before_script: "rubocop"
...
   srcclr: true
 after_success: "coveralls"
+
+jobs:
+  include:
+    - stage: 'Integration tests'
+      env: SDK=ruby
+      language: python
+      before_install:
+        - "pip install awscli"
```
mikeng13: I still think semantically, `pip install awscli` should be part of the `install` step and the cpoying of the artifacts should be in `before_script`

PR #152, opened 2019-01-08
feat(Audience Evaluation): Audience Logging

rashidsp:
Summary
-------
This adds logging to audience evaluation. 

Test plan
---------
Unit tests written in 
- spec/audience_spec.rb
- spec/custom_attribute_condition_evaluator_spec.rb

Issues
------
- OASIS-3852


lib/optimizely/custom_attribute_condition_evaluator.rb
```diff
@@ -77,9 +101,25 @@ def exact_evaluator(condition)
         return true if condition_value.to_f == user_provided_value.to_f
       end
 
-      return nil if !value_valid_for_exact_conditions?(user_provided_value) ||
-                    !value_valid_for_exact_conditions?(condition_value) ||
-                    !Helpers::Validator.same_types?(condition_value, user_provided_value)
+      unless value_valid_for_exact_conditions?(user_provided_value)
+        @logger.log(
+          Logger::WARN,
+          "Audience condition '#{condition}' evaluated as UNKNOWN because the value for user " \
```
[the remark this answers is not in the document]
rashidsp: This could be for logger improvement at SDK level.

lib/optimizely/custom_attribute_condition_evaluator.rb
```diff
@@ -73,13 +111,48 @@ def exact_evaluator(condition)
...
       user_provided_value = @user_attributes[condition['name']]
 
-      if user_provided_value.is_a?(Numeric) && condition_value.is_a?(Numeric)
-        return true if condition_value.to_f == user_provided_value.to_f
+      # Return true when 2 == 2.0
+      if condition_value.is_a?(Numeric) && user_provided_value.is_a?(Numeric) &&
+         Helpers::Validator.finite_number?(condition_value) &&
+         Helpers::Validator.finite_number?(user_provided_value)
+
+        return true if user_provided_value == condition_value
```
nchilada: Do we need to do some kind of special comparison instead of `user_provided_value == condition_value` in order to account for cases in which `user_provided_value` and `condition_value` are of different numeric types?

If not, I think it'd be cleaner to rely on the equality check that already occurs at the end of this function. At the very least, this would be consistent with most of our other SDKs. In exchange, we'd have to modify `Helpers::Validator.same_types?` to return true in the case of different numeric types.

lib/optimizely/custom_attribute_condition_evaluator.rb
```diff
@@ -133,20 +204,78 @@ def substring_evaluator(condition)
...
+        return nil
+      end
+
+      unless condition_value.is_a?(String)
+        @logger.log(
+          Logger::WARN,
+          format(Helpers::Constants::AUDIENCE_EVALUATION_LOGS['UNKNOWN_CONDITION_VALUE'], condition)
+        )
+        return nil
+      end
```
nchilada: Nit: would be slightly better (and certainly more consistent with other SDKs) to reorder the above two blocks. If the `condition_value` isn't a string, that'd be indicative of a broader problem that might be resolved in a future SDK, where it might actually be legitimate for the `user_provided_value` to be a non-string.

(I know we check for non-nil `user_provided_value` before we get to either of these blocks, but at least that helps us increase code re-use...)

spec/audience_spec.rb
```diff
@@ -193,4 +195,130 @@
...
+      'browser_type' => 5.5
+    }
+    experiment['audienceIds'] = %w[11110]
+
+    expect(Optimizely::Audience.user_in_experiment?(config,
+                                                    experiment,
+                                                    user_attributes)).to be false
+    expect(spy_logger).to have_received(:log).once.with(
+      Logger::DEBUG,
+      "Evaluating audiences for experiment 'test_experiment_with_audience': #{experiment['audienceIds']}."
```
nchilada: Can we avoid string interpolation in these tests? It'd be helpful to show what the final strings look like, especially since we're relying on Ruby's string representation of the underlying hash.

lib/optimizely/custom_attribute_condition_evaluator.rb
aliabbasrizvi: +1. It will be much cleaner that way.

lib/optimizely/helpers/validator.rb
```diff
@@ -153,9 +153,11 @@ def boolean?(value)
       def same_types?(value_1, value_2)
         # Returns true if given values are of same types.
         #         false otherwise.
+        # Numeric values are considered as same type.
+
+        return true if value_1.is_a?(Numeric) && value_2.is_a?(Numeric)
```
nchilada: Looks like it's okay to do the numeric check before the boolean check, in Ruby. I've confirmed that `true` and `false` are not `Numeric`s.

lib/optimizely/helpers/validator.rb
```diff
@@ -153,9 +153,11 @@ def boolean?(value)
...
 
         return true if boolean?(value_1) && boolean?(value_2)
-        return true if value_1.is_a?(Integer) && value_2.is_a?(Integer)
```
nchilada: Thanks for changing this! I just did some final QA for the 3.0 release, and I'm sure this would have been a problem if we hadn't changed it. We do expect numbers and floats to be interchangeable for the purposes of audience targeting. 👌

PR #162, opened 2019-03-26
feat(api): Feature variable APIs return default variable value when featureEnabled property is false.

rashidsp:
## Summary
- Feature variable APIs now return default variable value when featureEnabled property is false.
- Updated rubocop configuration to fix Metrics/BlockNesting error.

## Test plan
- Added unit tests.

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -473,7 +473,7 @@ module Optimizely
           variation = decision['variation']
           variation_variable_usages = @config.variation_id_to_variable_usage_map[variation['id']]
           variable_id = variable['id']
-          if variation_variable_usages&.key?(variable_id)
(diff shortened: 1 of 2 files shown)


lib/optimizely.rb
```diff
@@ -473,7 +473,7 @@ def get_feature_variable_for_type(feature_flag_key, variable_key, variable_type,
           variation = decision['variation']
           variation_variable_usages = @config.variation_id_to_variable_usage_map[variation['id']]
           variable_id = variable['id']
-          if variation_variable_usages&.key?(variable_id)
+          if variation['featureEnabled'] && variation_variable_usages&.key?(variable_id)
```
aliabbasrizvi: I think the log message in the else clause is incorrect for this new scenario. It should be the log message in the else clause for `decision` (above). Please update accordingly.

PR #166, opened 2019-04-19
feat(decision-listener): Incorporated new decision notification listener changes.

rashidsp:
## Summary
- Incorporated new decision notification listener changes.

## Test plan
- Added unit tests for type `feature-test` and project config method.
- Updated unit tests.

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -144,11 +144,17 @@ module Optimizely
 
       variation_id = @decision_service.get_variation(experiment_key, user_id, attributes)
       variation = @config.get_variation_from_id(experiment_key, variation_id) unless variation_id.nil?
+      experiment = @config.get_experiment_from_key(experiment_key)
+      experiment_id = experiment['id'] if experiment
       variation_key = variation['key'] if variation
-
+      decision_notification_type = if @config.feature_experiment?(experiment_id)
+                                     Helpers::Constants::DECISION_NOTIFICATION_TYPES['FEATURE_TEST']
+                                   else
+                                     Helpers::Constants::DECISION_NOTIFICATION_TYPES['AB_TEST']
+                                   end
       @notification_center.send_notifications(
         NotificationCenter::NOTIFICATION_TYPES[:DECISION],
-        Helpers::Constants::DECISION_INFO_TYPES['EXPERIMENT'], user_id, (attributes || {}),
+        decision_notification_type, user_id, (attributes || {}),
         experiment_key: experiment_key,
         variation_key: variation_key
       )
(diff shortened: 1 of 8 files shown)


lib/optimizely.rb
```diff
@@ -137,6 +137,9 @@ def get_variation(experiment_key, user_id, attributes = nil)
         }, @logger, Logger::ERROR
       )
 
+      experiment = @config.get_experiment_from_key(experiment_key)
```
oakbani: @rashidsp Can we remove this check from here now? 
https://github.com/optimizely/ruby-sdk/blob/master/lib/optimizely/decision_service.rb#L60

lib/optimizely/project_config.rb
```diff
@@ -451,6 +460,16 @@ def get_rollout_from_id(rollout_id)
       nil
     end
 
+    def feature_experiment?(experiment_id)
+      # Determines if a given experiment ID belongs to any Feature.
```
aliabbasrizvi: nit. `Determines if a given experiment ID is related to a feature.` or `Determines if given experiment is a feature test.`

spec/project_config_spec.rb
```diff
@@ -59,6 +59,21 @@
         'test_event_not_running' => config_body['events'][3]
       }
 
+      expected_experiment_feature_map = {
+        '133331' => [
```
aliabbasrizvi: An experiment having 2+ feature flags associated with it is an unexpected state. This points to me that the data we are testing with is incorrect / unexpected. Experiment to feature relationship is a many to one mapping and not many to many mapping.

cc @mikeng13

spec/project_config_spec.rb
```diff
@@ -1069,4 +1085,18 @@
       expect(config.get_attribute_id('$opt_user_agent')).to eq('$opt_user_agent')
     end
   end
+
+  describe '#feature_experiment' do
+    let(:config) { Optimizely::ProjectConfig.new(config_body_JSON, logger, error_handler) }
+
+    it 'should return true if the experiment belongs to any feature' do
```
aliabbasrizvi: nit. `.... if the experiment is a feature test`

spec/project_config_spec.rb
```diff
@@ -1069,4 +1085,18 @@
...
+      experiment = config.get_experiment_from_key('test_experiment_double_feature')
+      expect(config.feature_experiment?(experiment['id'])).to eq(true)
+    end
+
+    it 'should return false if the experiment does not belong to any feature' do
```
aliabbasrizvi: nit. `.... if the experiment is not a feature test`

spec/project_config_spec.rb
mikeng13: I just checked and these are mutex experiments, for which we do allow multiple features to be associated. So I think this data representation is correct.

mikeng13: Just synced offline with @aliabbasrizvi and turns out I had it backwards. An experiment can only be associated to one feature at a time. We should definitely fix this test data as it would otherwise suggest an erroneous case.

PR #168, opened 2019-05-16
Fix(is_feature_enabled):  Added rollout experiment key map for onboarding datafile

rashidsp:
## Summary
- Updated rollout experiment mapping.
- Added rollout experiment with different ID and key

## Test plan
- Modified existing mapping tests and added new one.

## Issues
- OASIS-4722

### DIFF
--- a/lib/optimizely/project_config.rb
+++ b/lib/optimizely/project_config.rb
@@ -55,6 +55,7 @@ module Optimizely
     attr_reader :group_key_map
     attr_reader :rollout_id_map
     attr_reader :rollout_experiment_id_map
+    attr_reader :rollout_experiment_key_map
     attr_reader :variation_id_map
     attr_reader :variation_id_to_variable_usage_map
     attr_reader :variation_key_map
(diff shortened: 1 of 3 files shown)


lib/optimizely/project_config.rb
```diff
@@ -120,13 +121,15 @@ def initialize(datafile, logger, error_handler)
         end
       end
       @rollout_id_map = generate_key_map(@rollouts, 'id')
-      # split out the experiment id map for rollouts
+      # split out the experiment id and key map for rollouts
       @rollout_experiment_id_map = {}
+      @rollout_experiment_key_map = {}
```
msohailhussain: Now where `rollout_experiment_id_map` is being used in the code.

lib/optimizely/project_config.rb
```diff
@@ -120,13 +120,13 @@ def initialize(datafile, logger, error_handler)
...
-      # split out the experiment id map for rollouts
-      @rollout_experiment_id_map = {}
+      # split out the experiment key map for rollouts
+      @rollout_experiment_key_map = {}
       @rollout_id_map.each_value do |rollout|
         exps = rollout.fetch('experiments')
-        @rollout_experiment_id_map = @rollout_experiment_id_map.merge(generate_key_map(exps, 'id'))
+        @rollout_experiment_key_map = @rollout_experiment_key_map.merge(generate_key_map(exps, 'key'))
       end
-      @all_experiments = @experiment_key_map.merge(@rollout_experiment_id_map)
```
mikeng13: Wow this is a big mistake we didn't catch in the code review last time around

aliabbasrizvi: Luckily IDs and keys have been same, but if it had changed things would be really very bad.

PR #178, opened 2019-07-01
feat(datafile-management): Integrate static_config_manager with Optimizely.

rashidsp:
Summary
-------

- Integrated `StaticProjectConfigManager` with Optimizely
- Removed attr_reader `is_valid` and created it as an instance method.
- Moved datafile validation code into newly created `ConfigManagerUtils`.

Test plan
---------
-  Existing unit tests
- New Unit tests


lib/optimizely.rb
```diff
@@ -470,9 +472,14 @@ def get_feature_variable_integer(feature_flag_key, variable_key, user_id, attrib
       variable_value
     end
 
+    def is_valid
```
mikeng13: Why not continue using `is_valid` in all the API methods? Seems less repetitive than having all the methods check whether the config itself is valid. We should also check for nil here.

lib/optimizely/helpers/config_manager_utils.rb
```diff
@@ -0,0 +1,55 @@
...
+#    limitations under the License.
+#
+require 'optimizely/logger'
+require 'optimizely/config/datafile_project_config'
+module Optimizely
+  module Helpers
+    module ConfigManagerUtils
+      module_function
+
+      def create_project_config_from_datafile(datafile, logger, error_handler, skip_json_validation)
```
mikeng13: THis doesn't need to be a util method. It can be a static method on the DatafileProjectConfig since this is tightly coupled with that class.

lib/optimizely.rb
```diff
@@ -374,12 +367,14 @@ def get_enabled_features(user_id, attributes = nil)
...
-      unless @is_valid
-        @logger.log(Logger::ERROR, InvalidDatafileError.new('get_feature_variable_string').message)
+      config = project_config
+      unless config.is_a?(Optimizely::ProjectConfig)
+        @logger.log(Logger::ERROR, InvalidProjectConfigError.new('get_feature_variable_string').message)
         return nil
       end
 
       variable_value = get_feature_variable_for_type(
+        config,
```
mikeng13: Why pass it in to this method? Can't this method just reference it?

lib/optimizely.rb
```diff
@@ -109,8 +97,8 @@ def activate(experiment_key, user_id, attributes = nil)
       end
 
       # Create and dispatch impression event
-      experiment = @config.get_experiment_from_key(experiment_key)
-      send_impression(experiment, variation_key, user_id, attributes)
+      experiment = config.get_experiment_from_key(experiment_key)
+      send_impression(config, experiment, variation_key, user_id, attributes)
```
mikeng13: Does Ruby offer any mutex / locking mechanisms so we can avoid passing this around?

lib/optimizely.rb
```diff
@@ -486,6 +485,12 @@ def get_feature_variable_for_type(feature_flag_key, variable_key, variable_type,
       # Returns nil if the feature flag or variable or user ID is empty
       #             in case of variable type mismatch
 
+      caller_method = caller_locations.first.label
```
mikeng13: This seems a bit flaky. I'd rather keep the is_valid checks in the top level variable APIs for consistency since all the other public APIs have the same checks there.

lib/optimizely.rb
```diff
@@ -470,8 +428,49 @@ def get_feature_variable_integer(feature_flag_key, variable_key, user_id, attrib
...
+      # attributes - Hash representing user attributes.
+      # config - Instance of DatfileProjectConfig
+      #
+      # Returns [variation key] where visitor will be bucketed.
+      # Returns [nil] if experiment is not Running, if user is not in experiment, or if datafile is invalid.
+      experiment = config.get_experiment_from_key(experiment_key)
+      return nil if experiment.nil?
+
+      unless user_inputs_valid?(attributes)
+        @logger.log(Logger::INFO, "Not activating user '#{user_id}.")
```
mikeng13: This is used by both `get_variation` and `activate`

rashidsp: Just removed it; Already being used in `user_inputs_valid?`.

PR #179, opened 2019-07-05
feat(datafile management): Add Polling Manager and Config Update

rashidsp:
## Summary
- Introduces HTTP Project Config Manager.
- Integrated HTTP Config Manager and Notification Center in Optimizely.
- Introduce generic Async scheduler for executing periodic task in a separate thread.
- Added Config Update Notification.

## Test plan

- Existing and New unit tests should pass.

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -75,6 +75,8 @@ module Optimizely
         @logger.log(Logger::ERROR, e.message)
       end
 
+      @notification_center = notification_center.is_a?(Optimizely::NotificationCenter) ? notification_center : NotificationCenter.new(@logger, @error_handler)
+
       @config_manager = if config_manager.respond_to?(:get_config)
                           config_manager
                         elsif sdk_key
@@ -84,14 +86,13 @@ module Optimizely
                             logger: @logger,
                             error_handler: @error_handler,
                             skip_json_validation: skip_json_validation,
-                            notification_center: notification_center
+                            notification_center: @notification_center
                           )
                         else
                           StaticProjectConfigManager.new(datafile, @logger, @error_handler, skip_json_validation)
                         end
       @decision_service = DecisionService.new(@logger, @user_profile_service)
       @event_builder = EventBuilder.new(@logger)
-      @notification_center = notification_center.is_a?(Optimizely::NotificationCenter) ? notification_center : NotificationCenter.new(@logger, @error_handler)
     end
 
     # Buckets visitor and sends impression event to Optimizely.
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -46,8 +46,10 @@ module Optimizely
     #                 By default all exceptions will be suppressed.
     # @param user_profile_service - Optional component which provides methods to store and retreive user profiles.
     # @param skip_json_validation - Optional boolean param to skip JSON schema validation of the provided datafile.
+    # @param config_manager - Optional Responds to get_config.
+    # @param notification_center - Optional Instance of NotificationCenter.
 
-    def initialize(datafile, event_dispatcher = nil, logger = nil, error_handler = nil, skip_json_validation = false, user_profile_service = nil)
+    def initialize(datafile = nil, event_dispatcher = nil, logger = nil, error_handler = nil, skip_json_validation = false, user_profile_service = nil, config_manager = nil, notification_center = nil)
       @logger = logger || NoOpLogger.new
       @error_handler = error_handler || NoOpErrorHandler.new
       @event_dispatcher = event_dispatcher || EventDispatcher.new
@@ -60,10 +62,14 @@ module Optimizely
         @logger.log(Logger::ERROR, e.message)
       end
 
-      @config_manager = StaticProjectConfigManager.new(datafile, @logger, @error_handler, skip_json_validation)
+      @config_manager = if config_manager.respond_to?(:get_config)
+                          config_manager
+                        else
+                          StaticProjectConfigManager.new(datafile, @logger, @error_handler, skip_json_validation)
+                        end
       @decision_service = DecisionService.new(@logger, @user_profile_service)
       @event_builder = EventBuilder.new(@logger)
-      @notification_center = NotificationCenter.new(@logger, @error_handler)
+      @notification_center = notification_center.is_a?(Optimizely::NotificationCenter) ? notification_center : NotificationCenter.new(@logger, @error_handler)
     end
 
     # Buckets visitor and sends impression event to Optimizely.
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -17,6 +17,7 @@
 #
 require_relative 'optimizely/audience'
 require_relative 'optimizely/config/datafile_project_config'
+require_relative 'optimizely/config_manager/http_project_config_manager'
 require_relative 'optimizely/config_manager/static_project_config_manager'
 require_relative 'optimizely/decision_service'
 require_relative 'optimizely/error_handler'
@@ -46,10 +47,22 @@ module Optimizely
     #                 By default all exceptions will be suppressed.
     # @param user_profile_service - Optional component which provides methods to store and retreive user profiles.
     # @param skip_json_validation - Optional boolean param to skip JSON schema validation of the provided datafile.
+    # @params sdk_key - Optional string uniquely identifying the datafile corresponding to project and environment combination.
+    #                   Must provide at least one of datafile or sdk_key.
     # @param config_manager - Optional Responds to get_config.
     # @param notification_center - Optional Instance of NotificationCenter.
 
-    def initialize(datafile = nil, event_dispatcher = nil, logger = nil, error_handler = nil, skip_json_validation = false, user_profile_service = nil, config_manager = nil, notification_center = nil)
+    def initialize(
+      datafile = nil,
+      event_dispatcher = nil,
+      logger = nil,
+      error_handler = nil,
+      skip_json_validation = false,
+      user_profile_service = nil,
+      sdk_key = nil,
+      config_manager = nil,
+      notification_center = nil
+    )
       @logger = logger || NoOpLogger.new
       @error_handler = error_handler || NoOpErrorHandler.new
       @event_dispatcher = event_dispatcher || EventDispatcher.new
@@ -64,6 +77,15 @@ module Optimizely
 
       @config_manager = if config_manager.respond_to?(:get_config)
                           config_manager
+                        elsif sdk_key
+                          HTTPProjectConfigManager.new(
+                            sdk_key: sdk_key,
+                            datafile: datafile,
+                            logger: @logger,
+                            error_handler: @error_handler,
+                            skip_json_validation: skip_json_validation,
+                            notification_center: notification_center
+                          )
                         else
                           StaticProjectConfigManager.new(datafile, @logger, @error_handler, skip_json_validation)
                         end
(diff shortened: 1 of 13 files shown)


lib/optimizely.rb
```diff
@@ -46,8 +46,10 @@ class Project
     #                 By default all exceptions will be suppressed.
     # @param user_profile_service - Optional component which provides methods to store and retreive user profiles.
     # @param skip_json_validation - Optional boolean param to skip JSON schema validation of the provided datafile.
+    # @param config_manager - Optional Responds to get_config.
+    # @param notification_center - Optional Instance of NotificationCenter.
 
-    def initialize(datafile, event_dispatcher = nil, logger = nil, error_handler = nil, skip_json_validation = false, user_profile_service = nil)
+    def initialize(datafile = nil, event_dispatcher = nil, logger = nil, error_handler = nil, skip_json_validation = false, user_profile_service = nil, config_manager = nil, notification_center = nil)
```
mikeng13: We should be able to also pass in the SDK key to use for instantiating a new `HTTPProjectConfigManager`

lib/optimizely.rb
```diff
@@ -60,10 +62,14 @@ def initialize(datafile, event_dispatcher = nil, logger = nil, error_handler = n
         @logger.log(Logger::ERROR, e.message)
       end
 
-      @config_manager = StaticProjectConfigManager.new(datafile, @logger, @error_handler, skip_json_validation)
+      @config_manager = if config_manager.respond_to?(:get_config)
```
mikeng13: If `sdk_key` is passed in, then we create a `HTTPProjectConfigManager` with it. Note, we should also be able to create a `DatafileProjectConfig` with the hard-coded datafile and pass it to the `HTTPProjectConfigManager` to have as its initial instance

lib/optimizely/config_manager/async_scheduler.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      begin
+        @running = true
+        @thread = Thread.new { execution_wrapper(@callback) }
+      rescue StandardError => e
+        @logger.log(
+          Logger::ERROR,
+          "Couldn't create a new thread for async scheduler."
+        )
+        @logger.log(Logger::DEBUG, e.message)
```
mikeng13: Let's combine the custom message with the `e.message` and pass that into the ERROR log call instead of having two log calls here.

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -0,0 +1,214 @@
...
+require 'json'
+module Optimizely
+  class HTTPProjectConfigManager < ProjectConfigManager
+    # Config manager that polls for the datafile and updated ProjectConfig based on an update interval.
+
+    attr_reader :config
+
+    # Initialize config manager. One of sdk_key or url has to be set to be able to use.
+    #
+    # sdk_key - Optional string uniquely identifying the datafile.
```
mikeng13: Let's also note here that the `sdk_key` is required unless a `url` is passed in.

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -0,0 +1,214 @@
...
+    end
+
+    def stop!
+      @async_scheduler.stop!
+    end
+
+    def get_config
+      return @config if ready?
+
+      if @async_scheduler.running
```
mikeng13: Can you add docs here to explain what is going on? I might be reading it wrong but it seems like we block `get_config` if there is a sync operation going on, right? If possible, I'd like us to still allow the previous config to be retrieved while a datafile sync is happening and only lock when the sync is done and we are about to replace the old config with the new one

lib/optimizely/helpers/constants.rb
```diff
@@ -359,6 +359,22 @@ module Constants
         'FEATURE_TEST' => 'feature-test',
         'FEATURE_VARIABLE' => 'feature-variable'
       }.freeze
+
+      CONFIG_MANAGER = {
+        'BLOCKING_TIMEOUT' => 15,
```
mikeng13: Can you document what this is

spec/config_manager/http_project_config_manager_spec.rb
```diff
@@ -0,0 +1,304 @@
...
+        sdk_key: 'QBw9gFM8oTn7ogY9ANCC1z',
+        datafile: config_body_JSON,
+        polling_interval: 1,
+        auto_update: false,
+        notification_center: notification_center,
+        logger: spy_logger
+      )
+
+      # Sleep to ensure that new request is not scheduled.
+      sleep 3
```
mikeng13: Will this be prone to flakiness and/or incorrect behavior? What if it takes more than 3 secs to schedule the next request?

oakbani: Polling interval has been set to 1. So we can safely assume that it shouldn't take more than 3 seconds to schedule another request. We can increase it to 5 but that increases delay in overall unit test running time.

lib/optimizely/config_manager/http_project_config_manager.rb
oakbani: We only block for @block_timeout seconds if the config hasn't been initialized yet, and a background thread is running in process. Updated comments

spec/config_manager/http_project_config_manager_spec.rb
mikeng13: Ok, let's leave it at 3 for now.

lib/optimizely.rb
```diff
@@ -46,8 +47,22 @@ class Project
...
+    # @param notification_center - Optional Instance of NotificationCenter.
+
+    def initialize(
+      datafile = nil,
+      event_dispatcher = nil,
+      logger = nil,
+      error_handler = nil,
+      skip_json_validation = false,
+      user_profile_service = nil,
+      sdk_key = nil,
```
msohailhussain: I am not in favor of adding all params. 
Either `sdk_key` or `config_manager` with `notification_center` should be enough. 
In case you add `sdk_key` then just add `config_manager` property to get access of PollingProjectConfigManager.

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -0,0 +1,223 @@
...
+      @async_scheduler.stop!
+    end
+
+    def get_config
+      # Get Project Config.
+
+      # Returns config immediately if the config has been initialized. When a hardcoded
+      # datafile is passed on init, config becomes ready immediately.
+
+      return @config if ready?
```
msohailhussain: This shouldn't directly return config. Please check this condition in C# or java. 
If manager is not started and get_config is called, it should always poll regardless config was already set or nil.

msohailhussain: https://github.com/optimizely/csharp-sdk/blob/master/OptimizelySDK/Config/PollingProjectConfigManager.cs#L108

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -0,0 +1,223 @@
...
+          return (url_template % sdk_key)
+        rescue
+          raise InvalidInputsError, "Invalid url_template #{url_template} provided."
+        end
+      end
+
+      url
+    end
+
+    def handle_response(response)
```
msohailhussain: Please bring this method just after fetch datafile. Easy to read the code.

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -0,0 +1,223 @@
...
+      if response.code == '304'
+        @logger.log(
+          Logger::DEBUG,
+          "Not updating config as datafile has not updated since #{@last_modified}."
+        )
+        return
+      end
+
+      @last_modified = response[Helpers::Constants::HTTP_HEADERS['LAST_MODIFIED']]
+      config = DatafileProjectConfig.create_project_config_from_datafile(response.body, @logger, @error_handler, @skip_json_validation) if response.body
```
msohailhussain: I am not sure but @mikeng13 would like to know your thoughts. 
`DatafileProjectConfig` itself representing `Datafile` association with `ProjectConfig`, the method name should be `create` only. :)

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -0,0 +1,223 @@
...
+    private
+
+    def polling_interval(polling_interval)
+      # Sets frequency at which datafile has to be polled and ProjectConfig updated.
+      #
+      # polling_interval - Time in seconds after which to update datafile.
+
+      # If polling interval is less than minimum allowed interval then set it to default update interval.
+
+      if polling_interval.to_i >= Helpers::Constants::CONFIG_MANAGER['MIN_UPDATE_INTERVAL']
```
msohailhussain: Please add MAX limit as well, what is supported in `sleep` method

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -0,0 +1,223 @@
...
+    )
+      @logger = logger || NoOpLogger.new
+      @error_handler = error_handler || NoOpErrorHandler.new
+      @datafile_url = get_datafile_url(sdk_key, url, url_template)
+      @polling_interval = nil
+      polling_interval(polling_interval)
+      blocking_timeout ||= Helpers::Constants::CONFIG_MANAGER['BLOCKING_TIMEOUT']
+      @blocking_timeout = blocking_timeout
+      @last_modified = nil
+      @async_scheduler = AsyncScheduler.new(method(:fetch_datafile), @polling_interval, auto_update, @logger)
```
msohailhussain: I would suggest break `fetch_datafile` into two pieces. 
1 - Only fetch datafile and parse it and return ProjectConfig. 
2 - NotificationCenter and ProjectConfig checking, either version is same or not should be done in SetConfig. 
These two methods should come into one method and that should be passed to `AsyncScheduler`

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -0,0 +1,223 @@
...
+      return if previous_revision == config.revision
+
+      unless ready?
+        @config = config
+        @mutex.synchronize { @resource.signal }
+      end
+
+      @config = config
+
+      @notification_center.send_notifications(NotificationCenter::NOTIFICATION_TYPES[:OPTIMIZELY_CONFIG_UPDATE])
```
msohailhussain: Please see comment above to refactor this method.

lib/optimizely/config_manager/async_scheduler.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    private
+
+    def execution_wrapper(callback)
+      # Executes the given callback periodically
+
+      loop do
+        begin
+          callback.call
```
msohailhussain: This will block execution until method is not finished? Is it possible to make it async.

msohailhussain: Also can you please check if you make it async, then the next scheduled call shouldn't overlap previous call and as soon as previous call is finished then check if the next call is due immediately then don't wait for sleep.

lib/optimizely.rb
mikeng13: I think if we want to make it a simpler initialization step, we can introduce the `OptimizelyFactory` class that we have in other SDKs that takes in less params and uses more defaults.

lib/optimizely/config_manager/async_scheduler.rb
mikeng13: I think this is a limitation of the language, where there are no async constructs out of the box and we will have to block the execution thread. I think this should be OK as long as we document it well.

lib/optimizely.rb
rashidsp: Done

PR #183, opened 2019-07-18
refact: Updates OptimizelyFactory methods.

rashidsp:
## Summary
- Revised Optimizely factory.
- Travis build will pass when PR: https://github.com/optimizely/ruby-sdk/pull/181 is merged!

### DIFF
--- a/lib/optimizely/optimizely_factory.rb
+++ b/lib/optimizely/optimizely_factory.rb
@@ -22,9 +22,9 @@ module Optimizely
     # Returns a new optimizely instance.
     #
     # @params sdk_key - Required String uniquely identifying the fallback datafile corresponding to project.
-    # @param fallback - Optional JSON string datafile.
-    def self.default_instance(sdk_key, fallback = nil)
-      Optimizely::Project.new(fallback, nil, nil, nil, nil, nil, sdk_key)
+    # @param fallback datafile - Optional JSON string datafile.
+    def self.default_instance(sdk_key, datafile = nil)
+      Optimizely::Project.new(datafile, nil, nil, nil, nil, nil, sdk_key)
     end
 
     # Returns a new optimizely instance.
@@ -37,7 +37,7 @@ module Optimizely
     # Returns a new optimizely instance.
     #
     # @params sdk_key - Required String uniquely identifying the datafile corresponding to project.
-    # @param fallback - Optional JSON string datafile.
+    # @param fallback datafile - Optional JSON string datafile.
     # @param event_dispatcher - Optional EventDispatcherInterface Provides a dispatch_event method which if given a URL and params sends a request to it.
     # @param logger - Optional LoggerInterface Provides a log method to log messages. By default nothing would be logged.
     # @param error_handler - Optional ErrorHandlerInterface which provides a handle_error method to handle exceptions.
@@ -48,7 +48,7 @@ module Optimizely
     # @param notification_center - Optional Instance of NotificationCenter.
     def self.custom_instance(
       sdk_key,
-      fallback = nil,
+      datafile = nil,
       event_dispatcher = nil,
       logger = nil,
       error_handler = nil,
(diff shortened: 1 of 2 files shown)


lib/optimizely/optimizely_factory.rb
```diff
@@ -19,37 +19,82 @@
...
 
-    def self.create_default_instance_with_sdk_key_and_datafile(sdk_key, datafile = nil)
-      Optimizely::Project.new(datafile, nil, nil, nil, nil, nil, sdk_key)
-    end
-
-    def self.create_default_instance_with_config_manager(config_manager)
+    # Returns a new optimizely instance.
+    #
+    # @param config_manager - Required ConfigManagerInterface Responds to get_config.
+    def self.default_instance_with_manager(config_manager)
```
mikeng13: With config_manager

lib/optimizely/optimizely_factory.rb
```diff
@@ -19,37 +19,82 @@
...
+    #
+    # @param config_manager - Required ConfigManagerInterface Responds to get_config.
+    # @param event_dispatcher - Optional EventDispatcherInterface Provides a dispatch_event method which if given a URL and params sends a request to it.
+    # @param logger - Optional LoggerInterface which provides a log method to log messages. By default nothing would be logged.
+    # @param error_handler - Optional ErrorHandlerInterface which provides a handle_error method to handle exceptions.
+    #                 By default all exceptions will be suppressed.
+    # @param user_profile_service - Optional UserProfileServiceInterface which provides methods to store and retreive user profiles.
+    # @param skip_json_validation - Optional Boolean param to skip JSON schema validation of the provided datafile.
+    # @param notification_center - Optional Instance of NotificationCenter.
+    def self.custom_instance_with_manager(
```
mikeng13: Maybe this is not needed. Let's keep it out until a customer asks for it. I don't see too much value in supporting this.

lib/optimizely/optimizely_factory.rb
```diff
@@ -19,38 +19,51 @@
...
+    def self.custom_instance(
+      sdk_key,
+      fallback = nil,
       event_dispatcher = nil,
       logger = nil,
       error_handler = nil,
       skip_json_validation = false,
       user_profile_service = nil,
-      sdk_key = nil,
-      config_manager = nil,
```
mikeng13: We should keep this config manager and pass it along to the constructor. We just shouldn't make it mandatory, default it to nil

lib/optimizely/optimizely_factory.rb
```diff
@@ -19,31 +19,46 @@
...
+    # @param logger - Optional LoggerInterface Provides a log method to log messages. By default nothing would be logged.
+    # @param error_handler - Optional ErrorHandlerInterface which provides a handle_error method to handle exceptions.
+    #                 By default all exceptions will be suppressed.
+    # @param skip_json_validation - Optional Boolean param to skip JSON schema validation of the provided datafile.
+    # @param user_profile_service - Optional UserProfileServiceInterface Provides methods to store and retreive user profiles.
+    # @param config_manager - Optional ConfigManagerInterface Responds to get_config.
+    # @param notification_center - Optional Instance of NotificationCenter.
+    def self.custom_instance(
+      sdk_key,
+      fallback = nil,
```
mikeng13: Externally, this should still be called `datafile`. Internally, we can call it fallback.

msohailhussain: fallback is coming from java-sdk? do u still want to change it? 
https://github.com/optimizely/java-sdk/blob/master/core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java#L152

mikeng13: Hmm, yeah this is a named parameter so I think we should just call it `datafile` and in the doc param we can call it the fallback datafile. I don't like the idea of introducing yet another name/concept for it.

rashidsp: @mikeng13 Done

PR #190, opened 2019-07-26
Implement get_feature_variable and create unit tests

brandonhudavid:
## Summary
- Implement `get_feature_variable` method in `optimizely.rb` to return value of variable with any type
- Changed `inputs_valid?` function in `validator.rb` to allow for `nil` variable type
- Create new unit tests in `project_spec.rb` to ensure functionality of `get_feature_variable`

Ruby is a dynamically-typed language, like [JavaScript](https://github.com/optimizely/javascript-sdk/pull/298), so it suffices to have a single method return the value of a feature variable rather than have a separate method for each possible feature variable type.

## Test plan
- Create unit tests in `project_spec.rb`
- Enable gherkin integration tests for Ruby in FSC suite and python-testapp

## Issues
- Fulfills one of the [SDK convenience APIs](https://docs.google.com/document/d/1MCsMioa4TNImryaDdOJJsZwL9j5zFvFIXODArHF3ORs/edit#) (design doc [here](https://docs.google.com/document/d/1CZ0iappDujwgQzYR8oiU9LRVzLho9jGQt2zaB1rpB48/edit#heading=h.6jynaot9cbnq))
- Passing Travis build: [here](https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/builds/121166632)

### DIFF
--- a/lib/optimizely/helpers/validator.rb
+++ b/lib/optimizely/helpers/validator.rb
@@ -132,6 +132,15 @@ module Optimizely
           variables.delete :user_id
         end
 
+        if variables.include? :variable_type
+          # Empty variable_type is a valid user ID.
+          unless variables[:variable_type].is_a?(String) || !variables[:variable_type]
+            is_valid = false
+            logger.log(level, "#{Constants::INPUT_VARIABLES['VARIABLE_TYPE']} is invalid")
+          end
+          variables.delete :variable_type
+        end
+
         variables.each do |key, value|
           next if value.is_a?(String) && !value.empty?
 
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -369,6 +369,32 @@ module Optimizely
       enabled_features
     end
 
+    # Get the value of the specified variable in the feature flag.
+    #
+    # @param feature_flag_key - String key of feature flag the variable belongs to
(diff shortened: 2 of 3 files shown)


lib/optimizely/helpers/validator.rb
```diff
@@ -132,6 +132,15 @@ def inputs_valid?(variables, logger = NoOpLogger.new, level = Logger::ERROR)
           variables.delete :user_id
         end
 
+        if variables.include? :variable_type
+          # Empty variable_type is a valid user ID.
```
mikeng13: huh? This can you explain this?

brandonhudavid: If `variables[:variable_type]` is anything other than a String or null value, it is invalid. Otherwise, it is valid, and so we remove `variable_type` from `variables`. That way, it does not go through the `value.is_a?(String) && !value.empty?` check on line 145.

msohailhussain:
@brandonhudavid Can you please resolve lint errors.

lib/optimizely/helpers/validator.rb
mikeng13: Got it. Thanks

PR #191, opened 2019-08-01
feat(eventProcessor): Add EventProcessor and BatchEventProcessor 

rashidsp:
## Summary
- Introduces an EventProcessor interface.
- Introduces a BatchEventProcessor
- Excluded `batch_event_processor.rb` from rubocop's Metrics/ParameterLists rule.

Buffering events within a queue before dispatching is an optimization that should prevent SDK implementations from exhausting resources while increasing throughput. This implementation relies on a `EventQueue` to buffer events received from one-to-many producers. A single consumer thread continuously polls from this queue to build a batch before emitting the batched `LogEvent`.

## Test Plan
- Added unit tests

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -21,14 +21,14 @@ Lint/LiteralAsCondition:
 Metrics/ParameterLists:
   Max: 6
   Exclude:
-    - 'lib/optimizely/config_manager/http_project_config_manager.rb'
     - 'lib/optimizely.rb'
-    - 'lib/optimizely/optimizely_factory.rb'
-    - 'lib/optimizely/event/entity/impression_event.rb'
-    - 'lib/optimizely/event/entity/snapshot_event.rb'
+    - 'lib/optimizely/config_manager/http_project_config_manager.rb'
+    - 'lib/optimizely/event/batch_event_processor.rb'
     - 'lib/optimizely/event/entity/conversion_event.rb'
     - 'lib/optimizely/event/entity/event_context.rb'
-    - 'lib/optimizely/event/batch_event_processor.rb'
+    - 'lib/optimizely/event/entity/impression_event.rb'
+    - 'lib/optimizely/event/entity/snapshot_event.rb'
+    - 'lib/optimizely/optimizely_factory.rb'
 
 Naming/AccessorMethodName:
   Exclude:
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -28,6 +28,7 @@ Metrics/ParameterLists:
     - 'lib/optimizely/event/entity/snapshot_event.rb'
     - 'lib/optimizely/event/entity/conversion_event.rb'
     - 'lib/optimizely/event/entity/event_context.rb'
+    - 'lib/optimizely/event/batch_event_processor.rb'
 
 Naming/AccessorMethodName:
   Exclude:
--- a/lib/optimizely/event/batch_event_processor.rb
+++ b/lib/optimizely/event/batch_event_processor.rb
@@ -25,34 +25,33 @@ module Optimizely
     attr_reader :event_queue
 
     DEFAULT_BATCH_SIZE = 10
-    DEFAULT_FLUSH_INTERVAL = 30_000
+    DEFAULT_BATCH_INTERVAL = 30_000
+    DEFAULT_QUEUE_CAPACITY = 1000
 
-    SHUTDOWN_SIGNAL = 'SHUTDOWN_SIGNAL'
     FLUSH_SIGNAL = 'FLUSH_SIGNAL'
+    SHUTDOWN_SIGNAL = 'SHUTDOWN_SIGNAL'
 
     def initialize(
       event_queue:,
       event_dispatcher:,
       batch_size:,
       flush_interval:,
-      start_by_default: false,
-      logger:
+      logger: nil
     )
-      @event_queue = event_queue || []
+      @event_queue = event_queue || SizedQueue.new(DEFAULT_QUEUE_CAPACITY)
       @event_dispatcher = event_dispatcher
       @batch_size = batch_size || DEFAULT_BATCH_SIZE
       @flush_interval = flush_interval || DEFAULT_BATCH_INTERVAL
-      @logger = logger
+      @logger = logger || NoOpLogger.new
       @mutex = Mutex.new
       @received = ConditionVariable.new
       @current_batch = []
-      @disposed = false
       @is_started = false
-      start! if start_by_default == true
+      start!
     end
 
     def start!
-      if (@is_started == true) && !@disposed
+      if @is_started == true
         @logger.log(Logger::WARN, 'Service already started.')
         return
       end
@@ -71,24 +70,24 @@ module Optimizely
     def process(user_event)
       @logger.log(Logger::DEBUG, "Received userEvent: #{user_event}")
 
-      if @disposed == true
+      unless @thread.alive?
         @logger.log(Logger::WARN, 'Executor shutdown, not accepting tasks.')
         return
       end
 
-      if @event_queue.include? user_event
-        @logger.log(Logger::WARN, 'Payload not accepted by the queue.')
-        return
-      end
-
       @mutex.synchronize do
-        @event_queue << user_event
+        begin
+          @event_queue << user_event
+        rescue Exception
+          @logger.log(Logger::WARN, 'Payload not accepted by the queue.')
+          return
+        end
         @received.signal
       end
     end
 
     def stop!
-      return if @disposed
+      return unless @thread.alive?
 
       @mutex.synchronize do
         @event_queue << SHUTDOWN_SIGNAL
@@ -97,13 +96,7 @@ module Optimizely
 
       @is_started = false
       @logger.log(Logger::WARN, 'Stopping scheduler.')
-      @thread.exit
-    end
-
-    def dispose
-      return if @disposed == true
-
-      @disposed = true
+      @thread.join
     end
 
     private
@@ -168,7 +161,7 @@ module Optimizely
       end
 
       # Reset the deadline if starting a new batch.
-      @flushing_interval_deadline = Helpers::DateTimeUtils.create_timestamp + @flush_interval if @current_batch.empty?
+      @flushing_interval_deadline = (Helpers::DateTimeUtils.create_timestamp + @flush_interval) if @current_batch.empty?
 
       @logger.log(Logger::DEBUG, "Adding user event: #{user_event.event['key']} to batch.")
       @current_batch << user_event
--- a/lib/optimizely/event/batch_event_processor.rb
+++ b/lib/optimizely/event/batch_event_processor.rb
@@ -16,13 +16,14 @@
 #    limitations under the License.
 #
 require_relative 'event_processor'
+require_relative '../helpers/validator'
 module Optimizely
   class BatchEventProcessor < EventProcessor
     # BatchEventProcessor is a batched implementation of the Interface EventProcessor.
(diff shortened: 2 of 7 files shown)


spec/event/batch_event_processor_spec.rb
```diff
@@ -0,0 +1,244 @@
...
+  MAX_DURATION_MS = 1000
+  TIMEOUT_INTERVAL_MS = 5000
+
+  before(:example) do
+    @event_queue = []
+    @event_dispatcher = Optimizely::EventDispatcher.new
+    allow(@event_dispatcher).to receive(:dispatch_event).with(instance_of(Optimizely::Event))
+  end
+
+  it 'return empty event queue event is processed' do
```
mnoman09: Can we also make sure if the dispatchEvent is getting called and logevent is correct?

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,202 @@
...
+      if should_split?(user_event)
+        flush_queue!
+        @current_batch = []
+      end
+
+      # Reset the deadline if starting a new batch.
+      @flushing_interval_deadline = Helpers::DateTimeUtils.create_timestamp + @flush_interval if @current_batch.empty?
+
+      @logger.log(Logger::DEBUG, "Adding user event: #{user_event.event['key']} to btach.")
+      @current_batch << user_event
```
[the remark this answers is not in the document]
rashidsp: We add mutex lock, when we add event into EventQueue!

.rubocop_todo.yml
```diff
@@ -28,6 +28,7 @@ Metrics/ParameterLists:
     - 'lib/optimizely/event/entity/snapshot_event.rb'
     - 'lib/optimizely/event/entity/conversion_event.rb'
     - 'lib/optimizely/event/entity/event_context.rb'
+    - 'lib/optimizely/event/batch_event_processor.rb'
```
msohailhussain: alphabetize order

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+
+    SHUTDOWN_SIGNAL = 'SHUTDOWN_SIGNAL'
+    FLUSH_SIGNAL = 'FLUSH_SIGNAL'
+
+    def initialize(
+      event_queue:,
+      event_dispatcher:,
+      batch_size:,
+      flush_interval:,
+      start_by_default: false,
```
msohailhussain: i don't think so it's really needed.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+      logger:
+    )
+      @event_queue = event_queue || []
+      @event_dispatcher = event_dispatcher
+      @batch_size = batch_size || DEFAULT_BATCH_SIZE
+      @flush_interval = flush_interval || DEFAULT_BATCH_INTERVAL
+      @logger = logger
```
msohailhussain: shouldn't it assign NoOpLogger if its nil

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+      @mutex = Mutex.new
+      @received = ConditionVariable.new
+      @current_batch = []
+      @disposed = false
```
msohailhussain: `disposed` is it also implemented in http_project_config_manager?

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+
+    def process(user_event)
+      @logger.log(Logger::DEBUG, "Received userEvent: #{user_event}")
+
+      if @disposed == true
+        @logger.log(Logger::WARN, 'Executor shutdown, not accepting tasks.')
+        return
+      end
+
+      if @event_queue.include? user_event
```
msohailhussain: what this condition mean?

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+      @current_batch = []
+    end
+
+    def add_to_batch(user_event)
+      if should_split?(user_event)
+        flush_queue!
+        @current_batch = []
+      end
+
+      # Reset the deadline if starting a new batch.
```
msohailhussain: Why this condition can't b in the `if`

lib/optimizely/event/batch_event_processor.rb
rashidsp: Updated for `DEFAULT_QUEUE_CAPACITY`

lib/optimizely/event/batch_event_processor.rb
rashidsp: Inline condition; Rubocop's auto-fix

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,193 @@
...
+module Optimizely
+  class BatchEventProcessor < EventProcessor
+    # BatchEventProcessor is a batched implementation of the Interface EventProcessor.
+    # Events passed to the BatchEventProcessor are immediately added to a EventQueue.
+    # The BatchEventProcessor maintains a single consumer thread that pulls events off of
+
+    attr_reader :event_queue
+
+    DEFAULT_BATCH_SIZE = 10
+    DEFAULT_BATCH_INTERVAL = 30_000
```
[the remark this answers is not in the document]
rashidsp: RuboCop's change, following ruby's style guide.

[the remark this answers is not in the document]
mikeng13: Thanks

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,193 @@
...
+        end
+
+        if item == FLUSH_SIGNAL
+          @logger.log(Logger::DEBUG, 'Received flush signal.')
+          flush_queue!
+          next
+        end
+
+        if item.is_a? Optimizely::UserEvent
+          @logger.log(Logger::DEBUG, "Received add to batch signal. with event: #{item.event['key']}.")
```
mikeng13: Let's not log here as it would be extremely noisy

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,193 @@
...
+      @logger.log(Logger::WARN, 'Stopping scheduler.')
+      @thread.exit
+    end
+
+    private
+
+    def run
+      loop do
+        if Helpers::DateTimeUtils.create_timestamp > @flushing_interval_deadline
+          @logger.log(
```
mikeng13: I think @msohailhussain mentioned that during testing this was very verbose as it was logged many times. If he is removing from C# SDK it should be removed here as well.

lib/optimizely/event/batch_event_processor.rb
msohailhussain: I think this is not noisy. because when any impression conversion event occurs, then it will be logged, which is OK. I may remove, if you still think it's noisy.

lib/optimizely/event/batch_event_processor.rb
msohailhussain: That's not the place, it's under this condition, am going to remove. 
`        if item.nil?
          @logger.log(Logger::DEBUG, 'Empty item, sleeping for 50ms.')
          sleep(0.05)
          next
        end`

lib/optimizely/event/batch_event_processor.rb
mikeng13: I still think it is noisy. Let's leave it out unless someone actually requests for this level of verbosity.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
+# frozen_string_literal: true
+
```
aliabbasrizvi: nit. add #

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+require_relative 'event_processor'
+require_relative '../helpers/validator'
+module Optimizely
+  class BatchEventProcessor < EventProcessor
+    # BatchEventProcessor is a batched implementation of the Interface EventProcessor.
+    # Events passed to the BatchEventProcessor are immediately added to a EventQueue.
+    # The BatchEventProcessor maintains a single consumer thread that pulls events off of
```
aliabbasrizvi: This sentence seems incomplete.

lib/optimizely/event/batch_event_processor.rb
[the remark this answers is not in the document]
aliabbasrizvi: Please add a comment about what the unit of this number is.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+      batch_size: DEFAULT_BATCH_SIZE,
+      flush_interval: DEFAULT_BATCH_INTERVAL,
+      logger: NoOpLogger.new
+    )
+      @event_queue = event_queue
+      @event_dispatcher = event_dispatcher
+      @batch_size = if (batch_size.is_a? Integer) && positive_number?(batch_size)
+                      batch_size
+                    else
+                      DEFAULT_BATCH_SIZE
```
aliabbasrizvi: A message should be logger here that `batch_size` is being set to default value.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+    end
+
+    def add_to_batch(user_event)
+      if should_split?(user_event)
+        flush_queue!
+        @current_batch = []
+      end
+
+      # Reset the deadline if starting a new batch.
+      @flushing_interval_deadline = (Helpers::DateTimeUtils.create_timestamp + @flush_interval) if @current_batch.empty?
```
aliabbasrizvi: If I understand correctly, `create_timestamp` creates time in local timezone: https://github.com/optimizely/ruby-sdk/blob/master/lib/optimizely/helpers/date_time_utils.rb#L23-L26

Shouldn't that time be in UTC?

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -0,0 +1,200 @@
...
+
+      current_context = @current_batch.last.event_context
+
+      new_context = user_event.event_context
+      # Revisions should match
+      unless current_context[:revision] == new_context[:revision]
+        @logger.log(Logger::DEBUG, 'Revisions mismatched: Flushing current batch.')
+        return true
+      end
+      # Projects should match
```
aliabbasrizvi: nit. New line above this.

lib/optimizely/event/batch_event_processor.rb
rashidsp: Rubocop's rule `Layout/EmptyLineAfterMagicComment`

lib/optimizely/event/batch_event_processor.rb
rashidsp: Done

lib/optimizely/event/batch_event_processor.rb
rashidsp: Done

lib/optimizely/event/batch_event_processor.rb
rashidsp: Done

lib/optimizely/event/batch_event_processor.rb
rashidsp: Updated

PR #194, opened 2019-08-08
feat(integrateep): Integrate Event processor - WIP

msohailhussain:
## Summary

The "why", or other context.

## Issues
- "THING-1234" or "Fixes #123"

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -22,6 +22,9 @@ require_relative 'optimizely/config_manager/static_project_config_manager'
 require_relative 'optimizely/decision_service'
 require_relative 'optimizely/error_handler'
 require_relative 'optimizely/event_builder'
+require_relative 'optimizely/event/forwarding_event_processor'
+require_relative 'optimizely/event/event_factory'
+require_relative 'optimizely/event/user_event_factory'
 require_relative 'optimizely/event_dispatcher'
 require_relative 'optimizely/exceptions'
 require_relative 'optimizely/helpers/constants'
@@ -30,13 +33,12 @@ require_relative 'optimizely/helpers/validator'
 require_relative 'optimizely/helpers/variable_type'
 require_relative 'optimizely/logger'
 require_relative 'optimizely/notification_center'
-
 module Optimizely
   class Project
     attr_reader :notification_center
     # @api no-doc
-    attr_reader :config_manager, :decision_service, :error_handler,
-                :event_builder, :event_dispatcher, :logger
+    attr_reader :config_manager, :decision_service, :error_handler, :event_dispatcher,
+                :event_processor, :logger, :stopped
 
     # Constructor for Projects.
     #
@@ -61,7 +63,8 @@ module Optimizely
       user_profile_service = nil,
       sdk_key = nil,
       config_manager = nil,
-      notification_center = nil
+      notification_center = nil,
+      event_processor = nil
     )
       @logger = logger || NoOpLogger.new
       @error_handler = error_handler || NoOpErrorHandler.new
@@ -92,7 +95,13 @@ module Optimizely
                           StaticProjectConfigManager.new(datafile, @logger, @error_handler, skip_json_validation)
                         end
       @decision_service = DecisionService.new(@logger, @user_profile_service)
-      @event_builder = EventBuilder.new(@logger)
+
+      @event_processor = if event_processor.respond_to?(:process)
+                           event_processor
+                         else
+                           ForwardingEventProcessor.new(@event_dispatcher, @logger, @notification_center)
+                         end
+      # @event_builder = EventBuilder.new(@logger)
     end
 
     # Buckets visitor and sends impression event to Optimizely.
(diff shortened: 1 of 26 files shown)


lib/optimizely.rb
```diff
@@ -92,7 +95,13 @@ def initialize(
                           StaticProjectConfigManager.new(datafile, @logger, @error_handler, skip_json_validation)
                         end
       @decision_service = DecisionService.new(@logger, @user_profile_service)
-      @event_builder = EventBuilder.new(@logger)
+
+      @event_processor = if event_processor.respond_to?(:process)
```
msohailhussain: what does it mean `respond_to` ?

lib/optimizely.rb
```diff
@@ -92,7 +95,13 @@ def initialize(
...
+                           event_processor
+                         else
+                           ForwardingEventProcessor.new(@event_dispatcher, @logger, @notification_center)
+                         end
+      # @event_builder = EventBuilder.new(@logger)
```
msohailhussain: remove it.

lib/optimizely.rb
```diff
@@ -243,20 +252,17 @@ def track(event_key, user_id, attributes = nil, event_tags = nil)
...
-      rescue => e
-        @logger.log(Logger::ERROR, "Unable to dispatch conversion event. Error: #{e}")
-      end
 
-      @notification_center.send_notifications(
-        NotificationCenter::NOTIFICATION_TYPES[:TRACK],
-        event_key, user_id, attributes, event_tags, conversion_event
-      )
+      if @notification_center.notification_count(NotificationCenter::NOTIFICATION_TYPES[:TRACK]).positive?
+        conversion_event = EventFactory.create_log_event(user_event, @logger)
```
msohailhussain: Please name it `log_event`

lib/optimizely.rb
```diff
@@ -507,6 +513,14 @@ def is_valid
       config.is_a?(Optimizely::ProjectConfig)
     end
 
+    def close
+      return if @stopped
+
+      @stopped = true
+      @config_manager.stop! if @config_manager.respond_to?(:stop!)
```
msohailhussain: how you are making sure, that stop property exists for both `event_processor` and `config_manager`

lib/optimizely.rb
```diff
@@ -692,15 +706,14 @@ def validate_instantiation_options
...
-        @logger.log(Logger::ERROR, "Unable to dispatch impression event. Error: #{e}")
-      end
+      user_event = UserEventFactory.create_impression_event(config, experiment, variation_id, user_id, attributes)
+      @event_processor.process(user_event)
+
+      return unless @notification_center.notification_count(NotificationCenter::NOTIFICATION_TYPES[:ACTIVATE]).positive?
+
+      @logger.log(Logger::INFO, "Activating user '#{user_id}' in experiment '#{experiment_key}'.")
       variation = config.get_variation_from_id(experiment_key, variation_id)
+      impression_event = EventFactory.create_log_event(user_event, @logger)
```
msohailhussain: `log_event`

lib/optimizely.rb
```diff
@@ -692,15 +706,14 @@ def validate_instantiation_options
...
       @notification_center.send_notifications(
         NotificationCenter::NOTIFICATION_TYPES[:ACTIVATE],
         experiment, user_id, attributes, variation, impression_event
```
msohailhussain: `log_event` change will be reflected here.

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -85,10 +85,13 @@ def ready?
 
     def start!
       @async_scheduler.start!
```
msohailhussain: Must check if it is stopped, then no need to start again.

lib/optimizely/event/forwarding_event_processor.rb
```diff
@@ -0,0 +1,44 @@
...
+#    limitations under the License.
+#
+require_relative 'event_processor'
+module Optimizely
+  class ForwardingEventProcessor < EventProcessor
+    # ForwardingEventProcessor is a basic transformation stage for converting
+    # the event batch into a LogEvent to be dispatched.
+    def initialize(event_dispatcher, logger, notification_center = nil)
+      @event_dispatcher = event_dispatcher
+      @logger = logger
```
msohailhussain: what if logger is nil.

lib/optimizely.rb
rashidsp: Checks instance method `process` exist, similarly we are checking for config_manager; `config_manager.respond_to?(:get_config)`

lib/optimizely.rb
rashidsp: `respond_to?(:stop!) ` ensures that the instance method `stop` exists!

mjc1283: I suggest `stop` should be treated as required for any `event_processor` or `config_manager` passed in. We could validate them in the constructor. When `stop` doesn't exist, we can log a message and ignore it.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -102,7 +102,7 @@ def stop!
         @received.signal
       end
 
-      @is_started = false
+      @started = false
```
mjc1283: What if `stop!` is called several times quickly? Would it be more correct this way, with checking `@started` first?

```ruby
def stop!
  return unless @started
  return unless @thread.alive? # not sure if this is still necessary?
  
  @logger.log(Logger::WARN, 'Stopping scheduler.')
  @started = false
  @mutex.synchronize do
    event_queue << SHUTDOWN_SIGNAL
    @received.signal
  end

  @thread.exit
end
```

lib/optimizely/event/forwarding_event_processor.rb
```diff
@@ -0,0 +1,38 @@
...
+    # ForwardingEventProcessor is a basic transformation stage for converting
+    # the event batch into a LogEvent to be dispatched.
+    def initialize(event_dispatcher, logger = nil)
+      @event_dispatcher = event_dispatcher
+      @logger = logger || NoOpLogger.new
+    end
+
+    def process(user_event)
+      log_event = Optimizely::EventFactory.create_log_event(user_event, @logger)
+
```
mjc1283: Is the log event notification implemented yet? I think it should be triggered here.

rashidsp: Implemented in PR: https://github.com/optimizely/ruby-sdk/pull/196

PR #197, opened 2019-08-29
feat(notification-center): Added LogEvent notification and integration.  

rashidsp:
## Summary
-  Introduced `LogEvent` notification in `notification_center.rb`.
- Integrated it in `batch_event_dispatcher.rb` and `forwarding_event_processor.rb`.
- Added `notification_count` in `notification_center.rb` and integrated into `Optimizely` for `notification types` _ACTIVATE_  and  _TRACK_.
## Test plan
- Added unit tests for `LogEvent` in `notification_center_spec.rb`, `batch_event_dispatcher_spec.rb` and `forwarding_event_processor_spec.rb`.
- Updated `Optimizely's` unit tests for `LogEvent` and `notification_count`.

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -102,7 +102,7 @@ module Optimizely
       @event_processor = if event_processor.respond_to?(:process)
                            event_processor
                          else
-                           ForwardingEventProcessor.new(@event_dispatcher, @logger)
+                           ForwardingEventProcessor.new(@event_dispatcher, @logger, @notification_center)
                          end
     end
 
@@ -258,11 +258,13 @@ module Optimizely
       @event_processor.process(user_event)
       @logger.log(Logger::INFO, "Tracking event '#{event_key}' for user '#{user_id}'.")
 
-      log_event = EventFactory.create_log_event(user_event, @logger)
-      @notification_center.send_notifications(
-        NotificationCenter::NOTIFICATION_TYPES[:TRACK],
-        event_key, user_id, attributes, event_tags, log_event
-      )
+      if @notification_center.notification_count(NotificationCenter::NOTIFICATION_TYPES[:TRACK]).positive?
+        log_event = EventFactory.create_log_event(user_event, @logger)
+        @notification_center.send_notifications(
+          NotificationCenter::NOTIFICATION_TYPES[:TRACK],
+          event_key, user_id, attributes, event_tags, log_event
+        )
+      end
       nil
     end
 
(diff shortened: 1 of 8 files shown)


lib/optimizely.rb
```diff
@@ -258,11 +258,13 @@ def track(event_key, user_id, attributes = nil, event_tags = nil)
       @event_processor.process(user_event)
       @logger.log(Logger::INFO, "Tracking event '#{event_key}' for user '#{user_id}'.")
 
-      log_event = EventFactory.create_log_event(user_event, @logger)
-      @notification_center.send_notifications(
-        NotificationCenter::NOTIFICATION_TYPES[:TRACK],
-        event_key, user_id, attributes, event_tags, log_event
-      )
+      if @notification_center.notification_count(NotificationCenter::NOTIFICATION_TYPES[:TRACK]).positive?
```
mjc1283: What do we gain by checking notification count, as opposed to just sending notifications regardless of count?

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -36,10 +36,11 @@ class BatchEventProcessor < EventProcessor
 
     def initialize(
       event_queue: SizedQueue.new(DEFAULT_QUEUE_CAPACITY),
-      event_dispatcher:,
+      event_dispatcher: Optimizely::EventDispatcher.new,
```
mjc1283: Why are we adding a default value here?

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -144,6 +151,16 @@ def run
 
         add_to_batch(item) if item.is_a? Optimizely::UserEvent
       end
+    rescue SignalException
```
mjc1283: This seems like an important change, but it isn't discussed in the summary. Could you explain this change? If this isn't related to LogEvent notification, then I think it should be in its own separate PR.

lib/optimizely.rb
rashidsp: This validates that user has already added the listener before sending the notification.
Please see java-sdk as a reference: https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/Optimizely.java#L304

lib/optimizely/event/batch_event_processor.rb
rashidsp: This is not related to LogEvent, therefore moved to PR: https://github.com/optimizely/ruby-sdk/pull/200

lib/optimizely/event/batch_event_processor.rb
rashidsp: Moved changes which are not related to LogEvent into PR: https://github.com/optimizely/ruby-sdk/pull/200.

PR #200, opened 2019-09-05
fix: Event processor fixes and updated config-manager polling interval

rashidsp:
## Summary
 - Set default Event Dispatcher In case of nil.
 - Log message for `flush_interval` validation.
 - Flush pending events when `stop` is called.
- `HTTPProjectConfigManager`  updates `polling_interval` validation. 
## Test plan
- Added/updated unit tests.

## Issues
- No issue filed.


msohailhussain:
@rashidsp Please update readme. valid duration is great than 0 and less than whatever maximum value is.

Issue #205, opened 2019-10-04
AsyncScheduler rescues errors but doesn't log the error


jasonkarns:
https://github.com/optimizely/ruby-sdk/blob/45e7e2b87041d9fc81d047c17c5c6429a0a731e5/lib/optimizely/config_manager/async_scheduler.rb#L78-L82

It rescues the error but doesn't log any of the error details that are necessary for tracing/debugging.

mikeng13:
Thanks for reporting! We will look into adding more details for tracing/debugging this path.

jasonkarns:
bumping comment: https://github.com/optimizely/ruby-sdk/pull/212#pullrequestreview-326232686

Tamara-Barum:
Closing as linked/Merged with 212 & 248 in May of 2020

PR #212, opened 2019-10-16
refact: Output error trace when callback execution fails 

oakbani:
## Summary
- Output error trace if callback execution fails.
- Added line ending in rubocop to avoid errors when working on windows. 

## Test plan
- Added unit test to assert logged error

## Issues
- OASIS-5466
https://github.com/optimizely/ruby-sdk/issues/205


lib/optimizely/config_manager/async_scheduler.rb
```diff
@@ -75,10 +75,10 @@ def execution_wrapper(callback)
       loop do
         begin
           callback.call
-        rescue
+        rescue StandardError => e
           @logger.log(
             Logger::ERROR,
-            'Something went wrong when running passed function.'
+            "Something went wrong when executing passed callback. #{e.message}"
```
jasonkarns: Shouldn't this be sent to the error handler so that users can handle these errors as desired? (In particular, without invoking the error handler, users can't record these exceptions to their error reporting services.)

PR #214, opened 2019-10-17
refact: Revise unit tests for http config manager

oakbani:
## Summary
Resolves:

    Spy logger warnings leaking into examples. 
    `raise_lifecycle_message': The use of doubles or partial doubles from rspec-mocks outside of the per-test lifecycle is not supported. (RSpec::Mocks::OutsideOfExampleError)

    Flaky unit tests due to compare timers. 

    Flaky unit test due to race condition. 

    Actual HTTP calls slowing down unit tests which can be mocked

The "why", or other context.

## Test plan
Existing and modified unit tests should continue to pass. 

## Issues
- OASIS-5521


spec/config_manager/http_project_config_manager_spec.rb
```diff
@@ -17,141 +17,166 @@
...
-      http_project_config_manager = Optimizely::HTTPProjectConfigManager.new(
-        sdk_key: 'QBw9gFM8oTn7ogY9ANCC1z',
+      @http_project_config_manager = Optimizely::HTTPProjectConfigManager.new(
+        sdk_key: 'valid_sdk_key',
         start_by_default: false
       )
       start = Time.now
-      http_project_config_manager.start!
-      until http_project_config_manager.ready?; end
+      @http_project_config_manager.start!
```
msohailhussain: can we remove timing difference logic and see the mocked method shouldn't get called. It's oK to me

oakbani: Found a nicer way to test blocking timeout part. Revised test.

PR #215, opened 2019-10-17
refact: Batch Event Processor and Unit Tests

oakbani:
## Summary
Addresses
_1. Event Batching thread can sleep anytime from 50 ms to 100ms. In other SDKs we strictly sleep for 50ms. This makes it inconsistent.
2. Mutex applied over Queue which already implements locking mechanism internally and is made for multithreaded access.
3. Unit tests filled up with varying sleep timers.
4. Unit tests throwing errors due to doubles/spys being passed to internal threads.
5. A message being logged in debug mode in a while loop. This fills up all logs and a client will certainly report it__

## Test plan
- Previous unit tests modified and new added.

## Issues
- OASIS-5495

### DIFF
--- a/lib/optimizely/event/batch_event_processor.rb
+++ b/lib/optimizely/event/batch_event_processor.rb
@@ -20,7 +20,7 @@ require_relative '../helpers/validator'
 module Optimizely
   class BatchEventProcessor < EventProcessor
     # BatchEventProcessor is a batched implementation of the Interface EventProcessor.
-    # Events passed to the BatchEventProcessor are immediately added to a EventQueue.
+    # Events passed to the BatchEventProcessor are immediately added to an EventQueue.
     # The BatchEventProcessor maintains a single consumer thread that pulls events off of
     # the BlockingQueue and buffers them for either a configured batch size or for a
     # maximum duration before the resulting LogEvent is sent to the NotificationCenter.
@@ -30,6 +30,7 @@ module Optimizely
     DEFAULT_BATCH_SIZE = 10
     DEFAULT_BATCH_INTERVAL = 30_000 # interval in milliseconds
     DEFAULT_QUEUE_CAPACITY = 1000
+    DEFAULT_TIMEOUT_INTERVAL = 5 # interval in seconds
 
     FLUSH_SIGNAL = 'FLUSH_SIGNAL'
     SHUTDOWN_SIGNAL = 'SHUTDOWN_SIGNAL'
@@ -69,6 +70,7 @@ module Optimizely
         return
       end
       @flushing_interval_deadline = Helpers::DateTimeUtils.create_timestamp + @flush_interval
+      @logger.log(Logger::INFO, 'Starting scheduler.')
       @thread = Thread.new { run }
       @started = true
     end
@@ -86,7 +88,7 @@ module Optimizely
       end
 
       begin
-        @event_queue << user_event
+        @event_queue.push(user_event, true)
       rescue Exception
         @logger.log(Logger::WARN, 'Payload not accepted by the queue.')
         return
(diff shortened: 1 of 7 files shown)


lib/optimizely/event/batch_event_processor.rb
```diff
@@ -90,56 +87,38 @@ def process(user_event)
...
-      @mutex.synchronize do
-        begin
-          @event_queue << user_event
-          @received.signal
-        rescue Exception
-          @logger.log(Logger::WARN, 'Payload not accepted by the queue.')
-          return
-        end
+      begin
+        @event_queue.push(user_event, true)
```
oakbani: Setting 2nd arg to true so that the main thread doesn't block and an exception is raised if the Queue is full. 
https://ruby-doc.org/core-2.3.0_preview1/SizedQueue.html#method-i-push

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -90,56 +87,38 @@ def process(user_event)
...
       return unless @started
 
-      @mutex.synchronize do
-        @event_queue << SHUTDOWN_SIGNAL
-        @received.signal
-      end
-
+      @logger.log(Logger::INFO, 'Stopping scheduler.')
+      @event_queue << SHUTDOWN_SIGNAL
+      @thread.join(DEFAULT_TIMEOUT_INTERVAL)
```
oakbani: @thread.exit would kill the inner thread and we may miss out events in the buffer. .join would instead wait for the thread to complete up to the given timeout.

lib/optimizely/event/batch_event_processor.rb
msohailhussain: can you please make sure in other SDKs, we are also getting queue size, but here we are just setting default queue size. Can you please add attribute to support that.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -90,56 +87,38 @@ def process(user_event)
...
     def run
       loop do
-        if Helpers::DateTimeUtils.create_timestamp > @flushing_interval_deadline
-          @logger.log(
-            Logger::DEBUG,
-            'Deadline exceeded flushing current batch.'
-          )
-          flush_queue!
-        end
+        flush_queue! if Helpers::DateTimeUtils.create_timestamp > @flushing_interval_deadline
```
msohailhussain: Can you pleae add logs in `flush_queue` while flushing.

oakbani: Done

lib/optimizely/event/batch_event_processor.rb
oakbani: We are already allowing user to pass it's own queue. Like we do in spec file, send our own SizedQueue with different size.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -90,56 +87,38 @@ def process(user_event)
...
 
-      @mutex.synchronize do
-        begin
-          @event_queue << user_event
-          @received.signal
-        rescue Exception
-          @logger.log(Logger::WARN, 'Payload not accepted by the queue.')
-          return
-        end
+      begin
```
jasonkarns: The begin/end wrapper is unnecessary.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -90,56 +87,38 @@ def process(user_event)
...
+        @event_queue.push(user_event, true)
+      rescue Exception
+        @logger.log(Logger::WARN, 'Payload not accepted by the queue.')
+        return
```
jasonkarns: This return redundant.

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -90,56 +87,38 @@ def process(user_event)
...
-        begin
-          @event_queue << user_event
-          @received.signal
-        rescue Exception
-          @logger.log(Logger::WARN, 'Payload not accepted by the queue.')
-          return
-        end
+      begin
+        @event_queue.push(user_event, true)
+      rescue Exception
```
jasonkarns: This should not be rescuing Exception. That will prevent the ruby process from safely exiting on `exit` and other base `Exception` subclasses. (should only be rescuing `StandardError`, which is the default behavior when no exception class is provided.)

Issue #213, opened 2019-10-16
Feature request: Notification center to conform to ActiveSupport notifications API


jasonkarns:
I have a couple concerns with the instrumentation implementation presently in use.

1) Currently, the Project's `notification_center` is being [type-checked against Optimizely::Notificiation center](https://github.com/optimizely/ruby-sdk/blob/99512e4c845a44024003be685d1331560c5792bd/lib/optimizely.rb#L83). This anti-pattern violates one of ruby's core idioms of duck typing. A notification center should only need to conform to the expected API. In particular, as long as the given object responds to `send_notifications`, then it should be considered a valid notification center.

2) Secondly, the existing NotificationCenter api does not conform to semi-standard notification APIs in Rails land. A de-facto standard API in Rails land is [`ActiveSupport::Notifications`](https://api.rubyonrails.org/classes/ActiveSupport/Notifications.html), of which there are many implementations. Conforming to this API would allow _existing_ instrumentation objects within a codebase to be passed to the Optimizely project to handle the new optimizely events.

3) ~~Lastly, the existing NotificationCenter only allows registration of `Method` listeners. This is completely abnormal ruby, as a normal ruby listener could be a block, proc, or lambda. Even better, it should be anything that responds to `call`, but at the very least, accepting block/proc/lambdas is table-stakes.~~ (addressed by #216 🎉 )

Tamara-Barum:
My apologies for the significant delay in response to this issue. The team is updating our processes and reviewing all open issues and PR's. If this Feature Request is still relevant, please add your suggestion to our idea portal here https://feedback.optimizely.com/ where it can be evaluated with the other ideas for the product.

PR #216, opened 2019-10-25
NotificationCenter should accept any Callable

jasonkarns:
## Summary

**The NotificationCenter learns to accept any callable object as the
  notification callback.**

It is uncommon in Ruby to require `Method` instances when any Callable
will suffice. Indeed, the idiomatic object for callbacks is a block.

With this change, the notification center can accept any callable, not
only `Method` instances; that is, any object that responds to `call`
(`Method`, lambdas, procs, or any object with a `.call` method)

If the callback parameter is omitted, a block may be provided instead.

## Issues

address item 3 of https://github.com/optimizely/ruby-sdk/issues/213

### DIFF
--- a/lib/optimizely/config_manager/http_project_config_manager.rb
+++ b/lib/optimizely/config_manager/http_project_config_manager.rb
@@ -70,14 +70,16 @@ module Optimizely
       @blocking_timeout = nil
       blocking_timeout(blocking_timeout)
       @last_modified = nil
-      @async_scheduler = AsyncScheduler.new(method(:fetch_datafile_config), @polling_interval, auto_update, @logger)
-      @async_scheduler.start! if start_by_default == true
-      @stopped = false
       @skip_json_validation = skip_json_validation
       @notification_center = notification_center.is_a?(Optimizely::NotificationCenter) ? notification_center : NotificationCenter.new(@logger, @error_handler)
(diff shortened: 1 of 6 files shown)


jasonkarns:
Not sure what's going on with Travis. The failure is occurring before the test step even begins. https://travis-ci.org/optimizely/ruby-sdk/jobs/602833534

jasonkarns:
> do you think it would be beneficial to provide a Proc.new {} test case as well (which also responds to :call)?

I can, but I don't think that additional test is valuable. We're already asserting that `Proc` instances can be provided (with the lambda test). It doesn't confirm anything that isn't already manifested in the tests and the `lambda` generation is more strict anyway (since the generated `Proc` checks arity).

Put another way, it's generally preferable to have only a single test fail for any given bug. Adding another test that fails for the same reason as an existing test doesn't provide value commensurate to its maintenance/carrying cost.

msohailhussain:
@jasonkarns can you please review this PR and merge it in this branch. It has lint fixes that should be resolved first.
https://github.com/jasonkarns/optimizely-ruby/pull/1

jasonkarns:
Merged! (feel free to make any future changes like that directly on the branch. I enabled "accept edits by maintainers" to facilitate this)

image

msohailhussain:
@jasonkarns Thanks.

PR #217, opened 2019-10-26
Place `Optimizely::Project` in optimizely/project.rb

jasonkarns:
## Summary
- Moves Optimizely::Project into the file corresponding to its name.
- Removes requires that aren't necessary (don't provide referenced constants within Project)
- Adds missing require
- Keeps "extraneous" requires in lib/optimizely.rb to make this change essentially a noop from behavior perspective.

Classes and modules should be defined in files that correspond to their name. This ensures that developers will know the right place to look because things will be in their conventional location.


brianlucas:
@jasonkarns This looks good, would you mind doing a rebase/remerge if it's not too much trouble and we can approve (assuming our tests pass)

oakbani:
Travis Link: https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/builds/139270452

jasonkarns:
too much drift.

PR #220, opened 2019-11-30
feat: get_optimizely_config API to get static experiments and features data

zashraf1985:
## Summary
Added a `get_optimizely_config` API which returns static experiments and features data.

## Test plan
1. Manually tested thoroughly
2. Added Unit tests
3. Full Stack Integration tests passed

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -33,6 +33,7 @@ require_relative 'optimizely/helpers/validator'
 require_relative 'optimizely/helpers/variable_type'
 require_relative 'optimizely/logger'
 require_relative 'optimizely/notification_center'
+require_relative 'optimizely/optimizely_config'
 
 module Optimizely
   class Project
@@ -521,6 +522,11 @@ module Optimizely
       @stopped = true
       @config_manager.stop! if @config_manager.respond_to?(:stop!)
       @event_processor.stop! if @event_processor.respond_to?(:stop!)
+    end
+
+    def get_optimizely_config
+      optimizely_config = OptimizelyConfig.new(project_config)
+      return optimizely_config.get_optimizely_config
     end
 
     private
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -524,7 +524,7 @@ module Optimizely
       @event_processor.stop! if @event_processor.respond_to?(:stop!)
     end
 
-    def get_optimizely_config # rubocop:disable Naming/AccessorMethodName
+    def get_optimizely_config
       # Get OptimizelyConfig object containing experiments and features data
       # Returns Object
       #
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -525,6 +525,42 @@ module Optimizely
     end
 
     def get_optimizely_config
+      # Get OptimizelyConfig object containing experiments and features data
+      # Returns Object
+      # 
+      # OptimizelyConfig Object Schema
+      # {
+      #   'experimentsMap' => {
+      #     '111111' => {
+      #       'id' => '111111',
+      #       'key' => 'my-fist-experiment'
+      #       'variationsMap' => {
+      #         '121212' => {
+      #           'id' => '121212',
+      #           'key' => 'variation_1',
+      #           'variablesMap' => {
+      #             '222222' => {
+      #               'id' => '222222',
+      #               'key' => 'age',
+      #               'type' => 'integer',
+      #               'value' => '0',
+      #             }
+      #           }
+      #         }
+      #       } 
+      #     }
+      #   },
+      #   'featuresMap' => {
+      #     '333333' => {
+      #       'id' => '333333',
+      #       'key' => 'awesome-feature',
+      #       'experimentsMap' => Object,
+      #       'variablesMap' => Object, 
+      #     } 
+      #   },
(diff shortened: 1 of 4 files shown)


lib/optimizely.rb
```diff
@@ -523,6 +524,47 @@ def close
...
     end
 
+    def get_optimizely_config # rubocop:disable Naming/AccessorMethodName
+      # Get OptimizelyConfig object containing experiments and features data
+      # Returns Object
+      #
+      # OptimizelyConfig Object Schema
+      # {
+      #   'experimentsMap' => {
+      #     '111111' => {
```
jaeopt: The examples are misleading as "id"-based map. Change them all to "key"-based map.

lib/optimizely.rb
```diff
@@ -523,6 +524,47 @@ def close
...
+      #       'id' => '333333',
+      #       'key' => 'awesome-feature',
+      #       'experimentsMap' => Object,
+      #       'variablesMap' => Object,
+      #     }
+      #   },
+      #   'revision' => '13',
+      # }
+      #
+      optimizely_config = OptimizelyConfig.new(project_config)
```
jaeopt: Add "unless is_valid" guard like other API calls, returning null when not ready.

lib/optimizely/optimizely_config.rb
```diff
@@ -0,0 +1,116 @@
...
+      }
+    end
+
+    private
+
+    def experiments_map
+      feature_variables_map = @project_config.feature_flags.reduce({}) do |result_map, feature|
+        result_map.update(feature['id'] => feature['variables'])
+      end
+      @project_config.experiments.reduce({}) do |experiments_map, experiment|
```
jaeopt: experiments should include experiments dedicated to "groups" as well. It looks like this list does not include them. Check it out.

zashraf1985: It does. Project Config already merged grouped experiments in `@experments` array. It contains all experiments except rollouts

lib/optimizely.rb
zashraf1985: Great catch. Fixed!. I copied them over from javascript-sdk. will fix it there too.

lib/optimizely.rb
zashraf1985: Done

lib/optimizely.rb
```diff
@@ -523,6 +524,52 @@ def close
       @event_processor.stop! if @event_processor.respond_to?(:stop!)
     end
 
+    def get_optimizely_config # rubocop:disable Naming/AccessorMethodName
```
oakbani: Instead of disabling it inline here. Please put in rubocop_todo.yml. Makes it easier to see all rule exceptions at one place. 
You can even use this utility to auto generate todo file for you. 

> rubocop --auto-gen-config

lib/optimizely/optimizely_config.rb
```diff
@@ -0,0 +1,116 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+module Optimizely
+  class OptimizelyConfig
```
oakbani: This is not a blocker as discussed offline, I would suggest that we should persist OptimizelyConfig response in this class. And each time config is called, it should expect a project config and we only generate new maps if the project ID  or revision has changed, otherwise we can return the same response. 

Datafile is unlikely to change so often. And we can save time from unnecessarily creating and merging so many maps to generate the same response again and again. Especially when the datafile is huge with many features/ experiments/ variables.

zashraf1985: Its a great suggestion and i remember we discussed it offline as well. I did it this way because its explicitly mentioned in the documentation that a new object will be generated every time. But i agree, this makes more sense to me as well. 
@jaeopt @mjc1283 .. What do you guys think about storing optimizelyconfig and creating a new one only if project config has a newer revision?

lib/optimizely.rb
zashraf1985: this particular rule can only be disabled for whole file using rubocop_todo which i think is not a good idea. What do you suggest ?

oakbani: I see. I would choose to keep in the todo file. But you may keep it this way.

msohailhussain: We shouldn't depend our code files on rubocop. Please add the rule in rubocop_todo.yml.

zashraf1985: Done

lib/optimizely/optimizely_config.rb
thomaszurkan-optimizely: @zashraf1985 @oakbani both Jae and Matt agree that we should add caching fo the OptimizelyConfig.  Can you update the PR?

zashraf1985: @thomaszurkan-optimizely You are abolutely right. But its also decided to do it in two separate PRs and merge them one by one in to master. So this should be considered as the first PR without caching. Once this is merged, we will create another one for caching.

PR #221, opened 2019-12-03
feat: Log and Handle Error for Event Dispatcher response

oakbani:
## Summary
- Allows Event Dispatcher to receive an optional logger and error handler
- Logs and calls Error Handler for Timeout or any other error raised. 
- Logs and calls Error Handler if response results into 4xx or 5xx status code. 

## Test plan
- Added new unit tests.

## Issues
- OASIS-5687

### DIFF
--- a/lib/optimizely/event/batch_event_processor.rb
+++ b/lib/optimizely/event/batch_event_processor.rb
@@ -37,7 +37,7 @@ module Optimizely
 
     def initialize(
       event_queue: SizedQueue.new(DEFAULT_QUEUE_CAPACITY),
-      event_dispatcher: Optimizely::EventDispatcher.new,
+      event_dispatcher: nil ,
       batch_size: DEFAULT_BATCH_SIZE,
       flush_interval: DEFAULT_BATCH_INTERVAL,
       logger: NoOpLogger.new,
@@ -45,7 +45,7 @@ module Optimizely
     )
       @event_queue = event_queue
       @logger = logger
-      @event_dispatcher = event_dispatcher
+      @event_dispatcher = event_dispatcher || EventDispatcher.new(logger: @logger)
       @batch_size = if (batch_size.is_a? Integer) && positive_number?(batch_size)
                       batch_size
                     else
--- a/lib/optimizely/event/batch_event_processor.rb
+++ b/lib/optimizely/event/batch_event_processor.rb
@@ -37,7 +37,7 @@ module Optimizely
 
     def initialize(
       event_queue: SizedQueue.new(DEFAULT_QUEUE_CAPACITY),
-      event_dispatcher: nil ,
+      event_dispatcher: nil,
       batch_size: DEFAULT_BATCH_SIZE,
       flush_interval: DEFAULT_BATCH_INTERVAL,
       logger: NoOpLogger.new,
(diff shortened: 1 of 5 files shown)


lib/optimizely/event_dispatcher.rb
```diff
@@ -28,26 +30,46 @@ class EventDispatcher
...
-        rescue Timeout::Error => e
-          return e
-        end
+        response = HTTParty.post(event.url,
+                                 body: event.params.to_json,
+                                 headers: event.headers,
+                                 timeout: REQUEST_TIMEOUT)
       end
+
+      error_msg = "Event failed to dispatch with response code: #{response.code}"
```
msohailhussain: Shouldn't it be in constants file?

oakbani: We don't follow this as a practice in the SDK. And I would prefer to keep this here as I don't expect it to be reused.

lib/optimizely/event_dispatcher.rb
```diff
@@ -28,26 +30,46 @@ class EventDispatcher
...
+
+      case response.code
+      when 400...500
+        @logger.log(Logger::ERROR, error_msg)
+        @error_handler.handle_error(HTTPCallError.new("HTTP Client Error: #{response.code}"))
```
jasonkarns: As a consumer, i would be super frustrated that I have to parse the error message to determine the actual http response code... I would request that the code be added as a property of the error type.

And as a stylistic point, I'll just note that instead of handling the error here, all this code could be a single `rescue => e`. the 400/500 responses could be raised, and the timeout/standard error rescues be combined to a single one that logs the error message and invokes the handler. I think that would simplify all this quite a bit.

lib/optimizely/event_dispatcher.rb
```diff
@@ -28,26 +30,46 @@ class EventDispatcher
     # @api constants
     REQUEST_TIMEOUT = 10
 
+    def initialize(logger: nil, error_handler: nil)
+      @logger = logger || NoOpLogger.new
+      @error_handler = error_handler || NoOpErrorHandler.new
```
jasonkarns: I'm confused why the keyword params are defaulted to nil, only to have that default overridden?

I think the idiomatic way to do this would be:
```
def initialize(logger: NoOpLogger.new, error_handler: NoOpErrorHandler.new)
  @logger = logger
  @error_handler = error_handler
end
```

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -37,15 +37,15 @@ class BatchEventProcessor < EventProcessor
...
+      event_dispatcher: nil,
       batch_size: DEFAULT_BATCH_SIZE,
       flush_interval: DEFAULT_BATCH_INTERVAL,
       logger: NoOpLogger.new,
       notification_center: nil
     )
       @event_queue = event_queue
       @logger = logger
-      @event_dispatcher = event_dispatcher
+      @event_dispatcher = event_dispatcher || EventDispatcher.new(logger: @logger)
```
jasonkarns: keyword args can be used within the method signature directly, there's no need to do all the manual defaulting within the body.

e.g.
```ruby
def initialize(
  logger: NoOpLogger.new,
  event_dispatcher: Optimizely::EventDispatcher.new(logger: logger)
)
```

PR #224, opened 2019-12-04
(fix): fix loop on batch event processor

thomaszurkan-optimizely:
## Summary
- The loop ends up running too often and not paying attention to flush interval.

## Test plan
- Needs unit tests

## Issues
- Batch event processor ends up in a tight always running loop when there are no events.

### DIFF
--- a/lib/optimizely/event/batch_event_processor.rb
+++ b/lib/optimizely/event/batch_event_processor.rb
@@ -107,14 +107,21 @@ module Optimizely
     private
 
     def run
+      @empty_count = 1
       loop do
-        flush_queue! if Helpers::DateTimeUtils.create_timestamp > @flushing_interval_deadline
+        if Helpers::DateTimeUtils.create_timestamp > @flushing_interval_deadline
+          flush_queue!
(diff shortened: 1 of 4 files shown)


lib/optimizely/event/batch_event_processor.rb
```diff
@@ -107,16 +108,26 @@ def stop!
...
-        flush_queue! if Helpers::DateTimeUtils.create_timestamp > @flushing_interval_deadline
+        if Helpers::DateTimeUtils.create_timestamp >= @flushing_interval_deadline
+          @logger.log(Logger::DEBUG, 'Deadline exceeded flushing current batch.')
+          flush_queue!
+          @flushing_interval_deadline = Helpers::DateTimeUtils.create_timestamp + @flush_interval
+          @use_pop = true if @nil_count > MAX_NIL_COUNT
+        end
 
-        item = @event_queue.pop if @event_queue.length.positive?
+        item = @event_queue.pop if @event_queue.length.positive? || @use_pop
```
mikeng13: Can you add a note here to explain what we are doing with the nil count?

PR #229, opened 2019-12-09
Should not be ignoring `rescue Exception` lint errors

jasonkarns:
It is almost always an error to be rescuing the base `Exception` class. See https://rubystyle.guide/#no-blind-rescues

"base" exceptions should be rescuing `StandardError`.

In cases where it truly is desired to rescue `Exception`, it should require the lint rule be explicitly disabled for that line, preferably with a comment explaining _why_. Disabling this rule by default is dangerous as it lets unintentional occurrences leak in. (as I believe has happened in this gem a number of times)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -45,8 +45,5 @@ Style/RescueStandardError:
 Style/SignalException:
   Enabled: false
 
-Lint/RescueException:
-  Enabled: false
-
 Layout/EndOfLine:
   EnforcedStyle: lf


jasonkarns:
Instances of rescued exceptions that should probably not be Exceptions (especially given the verbiage in the message indicates an expectation that Exception classes be caught, which is incorrect).

- https://github.com/optimizely/ruby-sdk/blob/d70733c6fdf9fb4f79364fd9ba84abdbe70f8462/lib/optimizely/event/batch_event_processor.rb#L152
- https://github.com/optimizely/ruby-sdk/blob/d70733c6fdf9fb4f79364fd9ba84abdbe70f8462/lib/optimizely/event/batch_event_processor.rb#L93

thomaszurkan-optimizely:
@jasonkarns I updated in this in https://github.com/optimizely/ruby-sdk/pull/232.  So, I'm closing this PR.  Thanks for the heads up!

PR #230, opened 2019-12-09
(fix): refactor for unicorn threads

thomaszurkan-optimizely:
## Summary
- Refactor to sleep remaining time before flush.  
- break out process events

## Test plan
Pass all existing tests

### DIFF
--- a/lib/optimizely/event/batch_event_processor.rb
+++ b/lib/optimizely/event/batch_event_processor.rb
@@ -123,7 +123,7 @@ module Optimizely
 
         add_to_batch(item) if item.is_a? Optimizely::UserEvent
       end
-      return true
+      true
     end
 
     def run
@@ -139,9 +139,9 @@ module Optimizely
 
         break unless process_events
 
-        interval = (Helpers::DateTimeUtils.create_timestamp - @flushing_interval_deadline)/1.0
+        interval = (Helpers::DateTimeUtils.create_timestamp - @flushing_interval_deadline) / 1.0
 
-        sleep interval if interval > 0
+        sleep interval if interval.positive?
       end
     rescue SignalException
       @logger.log(Logger::ERROR, 'Interrupted while processing buffer.')
--- a/lib/optimizely/event/batch_event_processor.rb
+++ b/lib/optimizely/event/batch_event_processor.rb
@@ -31,7 +31,6 @@ module Optimizely
     DEFAULT_BATCH_INTERVAL = 30_000 # interval in milliseconds
     DEFAULT_QUEUE_CAPACITY = 1000
     DEFAULT_TIMEOUT_INTERVAL = 5 # interval in seconds
-    MAX_NIL_COUNT = 3
 
     FLUSH_SIGNAL = 'FLUSH_SIGNAL'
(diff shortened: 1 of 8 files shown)


lib/optimizely/event/batch_event_processor.rb
```diff
@@ -107,36 +107,12 @@ def stop!
...
-          # this avoids to much spinning of the loop.
-          @nil_count += 1
-          next
-        end
-
-        # reset nil_count and use_pop if we have received an item.
-        @nil_count = 0
-        @use_pop = false
-
+    def process_events
```
mikeng13: nit: Would it be more apt to call this `process_queue` instead since you can have non-event items in there?

lib/optimizely/event/batch_event_processor.rb
```diff
@@ -147,15 +135,35 @@ def run
...
+        end
+
+        break unless process_queue
+
+        # what is the current interval to flush in seconds
+        interval = (@flushing_interval_deadline - Helpers::DateTimeUtils.create_timestamp) * 0.001
+
+        next unless interval.positive?
+
+        @wait_mutex.synchronize { @resource.wait(@wait_mutex, interval) }
```
oakbani: Can we run into a race condition where flush, stop or process is called from the main thread and hence the resource is signaled while we are processing the queue? And this thread waits later. In this scenario, the thread would wait up to the interval before processing any events. That may become an issue if the interval is long.

oakbani:
@thomaszurkan-optimizely  As per docs, the Queue class implements locking mechanism internally. Aren't we doing same as what event_queue.pop in blocking mode would do? That is suspend the thread.

lib/optimizely/event/batch_event_processor.rb
thomaszurkan-optimizely: Each of those elements adds an element to the queue. in the case of stop, the process exits, if we are not in the synchronize block and receive a signal, it is ignored.  So, this would have to happen during interval calculation time.   But, I could add a test to see if the queue length is positive before the wait.

lib/optimizely/event/forwarding_event_processor.rb
```diff
@@ -30,11 +30,15 @@ def process(user_event)
       log_event = Optimizely::EventFactory.create_log_event(user_event, @logger)
 
       begin
-        @event_dispatcher.dispatch_event(log_event)
-        @notification_center&.send_notifications(
-          NotificationCenter::NOTIFICATION_TYPES[:LOG_EVENT],
-          log_event
-        )
+        Thread.new do
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: Synchronously 1 at a time on the same thread can't scale and so no one will ever use it. However, we could make this configurable.

thomaszurkan-optimizely:
@oakbani the queue class does have a lock but no timeout.  so, if no other event came the queue would never be flushed on timeout interval.

lib/optimizely/event/forwarding_event_processor.rb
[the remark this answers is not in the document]
mikeng13: I think this makes sense, but does it make sense to do in a follow-up PR? Seems like it'll break a lot of tests. That way we can keep just the batching fix in this PR

PR #233, opened 2019-12-19
(fix): update the forwarder to use a thread

thomaszurkan-optimizely:
## Summary
- Update the event forwarder to use a thread instead of using the main thread.

## Test plan
- update tests

### DIFF
--- a/lib/optimizely/event/forwarding_event_processor.rb
+++ b/lib/optimizely/event/forwarding_event_processor.rb
@@ -30,11 +30,14 @@ module Optimizely
       log_event = Optimizely::EventFactory.create_log_event(user_event, @logger)
 
       begin
-        @event_dispatcher.dispatch_event(log_event)
-        @notification_center&.send_notifications(
-          NotificationCenter::NOTIFICATION_TYPES[:LOG_EVENT],
-          log_event
-        )
+        @notification_center&.send_notifications(NotificationCenter::NOTIFICATION_TYPES[:LOG_EVENT], log_event)
+        Thread.new do
+          begin
+            @event_dispatcher.dispatch_event(log_event)
+          rescue StandardError => e
+            @logger.log(Logger::ERROR, "Error dispatching event: #{log_event} #{e.message}.")
+          end
+        end
       rescue StandardError => e
         @logger.log(Logger::ERROR, "Error dispatching event: #{log_event} #{e.message}.")
       end
--- a/spec/project_spec.rb
+++ b/spec/project_spec.rb
@@ -187,6 +187,7 @@ describe 'Optimizely' do
       stub_request(:post, impression_log_url).with(query: params)
 
       expect(project_instance.activate('test_experiment', 'test_user')).to eq('control')
+      sleep 0.1
       expect(project_instance.event_dispatcher).to have_received(:dispatch_event).with(Optimizely::Event.new(:post, impression_log_url, params, post_headers)).once
       expect(project_instance.decision_service.bucketer).to have_received(:bucket).once
     end
@@ -205,6 +206,7 @@ describe 'Optimizely' do
       stub_request(:post, impression_log_url).with(query: params)
 
       expect(project_instance.activate('test_experiment', 'test_user')).to eq('control')
+      sleep 0.1
       expect(project_instance.event_dispatcher).to have_received(:dispatch_event).with(Optimizely::Event.new(:post, impression_log_url, params, post_headers)).once
     end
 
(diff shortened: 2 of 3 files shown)


lib/optimizely/event/forwarding_event_processor.rb
```diff
@@ -30,11 +30,14 @@ def process(user_event)
       log_event = Optimizely::EventFactory.create_log_event(user_event, @logger)
 
       begin
-        @event_dispatcher.dispatch_event(log_event)
-        @notification_center&.send_notifications(
-          NotificationCenter::NOTIFICATION_TYPES[:LOG_EVENT],
-          log_event
-        )
+        @notification_center&.send_notifications(NotificationCenter::NOTIFICATION_TYPES[:LOG_EVENT], log_event)
```
oakbani: Shouldn't we be only firing notification when the event has successfully dispatched?

lib/optimizely/event/forwarding_event_processor.rb
```diff
@@ -30,11 +30,14 @@ def process(user_event)
...
+        Thread.new do
+          begin
+            @event_dispatcher.dispatch_event(log_event)
+          rescue StandardError => e
```
oakbani: Can we run into second rescue when we have rescued here?

spec/project_spec.rb
```diff
@@ -492,10 +497,12 @@ class InvalidErrorHandler; end
       project_instance.decision_service.set_forced_variation(project_config, 'test_experiment_with_audience', 'test_user', 'variation_with_audience')
       variation_to_return = project_instance.decision_service.get_forced_variation(project_config, 'test_experiment', 'test_user')
       allow(project_instance.decision_service.bucketer).to receive(:bucket).and_return(variation_to_return)
+      sleep 0.1
```
oakbani: I am not sure if we need to put a sleep before **allow**.

spec/project_spec.rb
```diff
@@ -595,13 +602,16 @@ def callback(_args); end
         method(:callback)
       )
       variation_to_return = project_config.get_variation_from_id('test_experiment', '111128')
+      sleep 0.1
```
oakbani: I am not sure if we need to put a sleep before allow.

spec/project_spec.rb
```diff
@@ -741,18 +757,20 @@ def callback(_args); end
 
         expect(http_project_config_manager.config).not_to eq(nil)
         expect(project_instance.activate('checkout_flow_experiment', 'test_user')).not_to eq(nil)
+        sleep 0.5
```
oakbani: why sleeping for 0.5 here? I fear flakiness in  Unit Tests on Travis.

spec/project_spec.rb
```diff
@@ -2814,7 +2852,7 @@ def callback(_args); end
       project_instance = Optimizely::Project.new(nil, nil, nil, nil, true, nil, nil, config_manager, nil, event_processor)
 
       expect(config_manager.stopped).to be false
-      expect(event_processor.started).to be true
+      expect(event_processor.started).to be false
```
oakbani: doesn't belong to this PR?

lib/optimizely/event/forwarding_event_processor.rb
oakbani: Also, why not also fire notification in the thread as well. this will allow us to quickly free up our main thread if the client has registered some time-taking callback in the LogEvent listener. And, if we do this, we can simply wait for the callback to get called in our unit tests rather than putting it to sleep before assertions.

thomaszurkan-optimizely: Yeah, I was debating back and forth on that.  Originally, it was in the thread.  However, all the other SDKs fire the event before dispatch.  So, I thought this was ok.  I can put it back in the thread.  it might be better for the notification listener to not be on the main thread.

lib/optimizely/event/forwarding_event_processor.rb
thomaszurkan-optimizely: The first rescue is for the event dispatch failure.  really just for logging.  the second is for thread failure.

spec/project_spec.rb
thomaszurkan-optimizely: there is a get_forced _varitation before that.  the sleep is to give the thread a chance to run.

spec/project_spec.rb
thomaszurkan-optimizely: the thread will not have a chance to run without the sleep.

aliabbasrizvi:
Closing due to inactivity. Please re-open if needed.

PR #235, opened 2019-12-31
feat: Added caching for optimizelyConfig object

zashraf1985:
## Summary
Updated the `OptimizelyConfig` implementation to always return cached object. `OptimizelyConfig` object will only be updated when new `ProjectConfig` is received.

## Test Plan
Added Unit Tests

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -566,8 +566,13 @@ module Optimizely
         return nil
       end
 
-      optimizely_config = OptimizelyConfig.new(project_config)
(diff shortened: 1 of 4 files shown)


zashraf1985:
@mikeng13 @thomaszurkan-optimizely It would be great if one of you can take a look and approve too.

zashraf1985:
> Please update the header on the rb files you touched. copy right should be through 2020. :)

Good Catch @thomaszurkan-optimizely . Its done :)

PR #239, opened 2020-02-06
add spell check to markdown files

juancarlostong:
## Summary
- Adds spellchecker to *.md

This is what it looks like when it finds a spelling error (aborts build):
https://travis-ci.org/optimizely/ruby-sdk/jobs/647067279

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ jobs:
         email: false
     - stage: 'Lint markdown files'
       language: node_js
+      node_js: 12
       os: linux
       before_install: skip
       install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ jobs:
       install:
         - npm i -g markdown-spellcheck
       before_script:
-        - wget https://gist.githubusercontent.com/juancarlostong/dad02feeebc8763af35b4fdc717cf7a2/raw/29e085302a9418edbefba15d8f790fa76e0a89d5/.spelling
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -38,6 +38,18 @@ jobs:
...
         email: false
+    - stage: 'Lint markdown files'
+      language: node_js
+      node_js: 12
+      os: linux
+      before_install: skip
+      install:
+        - npm i -g markdown-spellcheck
+      before_script:
+        - wget --quiet https://gist.githubusercontent.com/juancarlostong/dad02feeebc8763af35b4fdc717cf7a2/raw/29e085302a9418edbefba15d8f790fa76e0a89d5/.spelling
```
juancarlostong: opting for a simple wget rather than the more expensive cloning of an entire repo. we only need this one file anyway and there's no secrets in it.

juancarlostong:
if this PR looks good, i will create similar PRs on all the remaining SDK repos

.travis.yml
mikeng13: Does it have to be node? Is there a more performant platform we can use? 

Also, I'd prefer if we can run this in parallel to the other stages so we don't have to wait for it

juancarlostong: it runs in parallel within the 'Lint markdown files' stage. in practice: 
Screen Shot 2020-02-07 at 10 36 26 AM

here you can see it finishes in 39 seconds, which is faster than the other job that ran in parallel (43s) meaning that it does not add to the existing build time

Issue #211, opened 2019-10-15
Feature request: eliminate HTTParty dependency


jasonkarns:
In a large Rails app, the number of dependencies can grow quite large. In such an app, the risk of unresolveable version conflicts between versions of a single dependency grows ever higher. As a library, it would be very helpful to consumers if the library's own dependencies were kept to a minimum. This would minimize or eliminate any such version conflicts.

Case in point, the HTTParty gem is a dependency that _could_ be removed. From what I can tell, there are only two small use-cases for HTTP in this gem: [event dispatcher](https://github.com/optimizely/ruby-sdk/blob/99512e4c845a44024003be685d1331560c5792bd/lib/optimizely/event_dispatcher.rb) and the [http config manager](https://github.com/optimizely/ruby-sdk/blob/99512e4c845a44024003be685d1331560c5792bd/lib/optimizely/config_manager/http_project_config_manager.rb). Neither object has that much extensive use of HTTP.

I would request that these classes just use [net/http](https://ruby-doc.org/stdlib-2.6.5/libdoc/net/http/rdoc/Net/HTTP.html) from the ruby standard library instead of pulling in a dependency.

Tamara-Barum:
Closing as linked/Merged with 252

PR #242, opened 2020-03-05
add ability to use http proxy when making web requests

### LINKED ISSUE #241: [feature request] Add ability for HTTP Config Manager to take proxy options
I propose adding the following options to HTTP Config manager:

| name | default | description |
| --- | --- | --- |
| proxy_host | nil | To perform a web request behind a proxy, this is required. This can be a domain or IP address of the proxy server.
| proxy_port | nil | port number that the proxy server uses for client connections. |
| proxy_username | nil | username for proxy auth. |
| proxy_password | nil | password for proxy auth. |

Happy to submit a PR to add this functionality.

rloomba:
## Summary
- add ability to use a proxy server when making http requests
- modify`HTTPProjectConfigManager` and `EventDispatcher` to take in proxy config options

I am unable to use this library as all my web requests must be routed through a web proxy.

[HTTParty proxy config documentation](https://github.com/jnunemaker/httparty/blob/master/lib/httparty.rb#L37-L40)

## Issues
- Fixes #241

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,8 @@ jobs:
       install:
         - npm i -g markdown-spellcheck
       before_script:
-        - wget --quiet https://raw.githubusercontent.com/optimizely/mdspell-config/master/.spelling
+        # todo: change branch to master once merged.
+        - wget --quiet https://raw.githubusercontent.com/optimizely/mdspell-config/oakbani/ignore-for-ruby/.spelling
       script:
         - mdspell -a -n -r --en-us '**/*.md'
       after_success: skip
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ jobs:
         - npm i -g markdown-spellcheck
       before_script:
         # todo: change branch to master once merged.
-        - wget --quiet https://raw.githubusercontent.com/optimizely/mdspell-config/oakbani/ignore-for-ruby/.spelling
+        - wget --quiet https://raw.githubusercontent.com/optimizely/mdspell-config/master/.spelling
       script:
         - mdspell -a -n -r --en-us '**/*.md'
       after_success: skip
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -14,7 +14,7 @@ Lint/HandleExceptions:
 # Offense count: 8
 # Configuration parameters: CountKeywordArgs.
 Metrics/ParameterLists:
-  Max: 13
+  Max: 14
 
 # Offense count: 2
 Naming/AccessorMethodName:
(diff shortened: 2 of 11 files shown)


rloomba:
@thomaszurkan-optimizely sounds good! Happy to do the refactor.

jasonkarns:
Does it make sense to further entrench the httparty gem via this configuration? As I opened in #211 , I think it makes sense to eliminate the gem dependency entirely and just depend on Ruby Core/StdLib alone. (ie, net/http or open-uri) Both of the stdlib tools respect HTTP Proxy env vars out of the box and would require no configuration at all.

rloomba:
I'm 100% onboard with #211.

I don't think we'll entrench this gem that much further If we extract the proxy config settings to a separate config object that @thomaszurkan-optimizely mentioned [above](https://github.com/optimizely/ruby-sdk/pull/242#pullrequestreview-394003898), couldn't that be used in whatever HTTP library we end up moving to?

I think respecting the HTTP Proxy env var is a good approach, but I also like having the option to explicitly config the proxy in application level code. What if we have multiple proxies that we'd like switch between?

@jasonkarns implementing #211 seems like a bigger chunk of work that I'm not sure I can lead, but  I'm happy to pitch in if there are some smaller pieces to be worked on or at least offer some code review.

jasonkarns:
> I think respecting the HTTP Proxy env var is a good approach, but I also like having the option to explicitly config the proxy in application level code. What if we have multiple proxies that we'd like switch between?

Good call. The trick will be ensuring that we distinguish the options being unset (to allow env vars to affect), from them being intentionally set to nil (to ignore env vars and disable the proxy).

thomaszurkan-optimizely:
Hi @rloomba,
We removed httparty and created a HttpUtils class to wrap our http calls.  Can you refactor your proxy settings here and we can merge it?  Otherwise, we will do it sometime in the next few sprints.  Thanks!

rloomba:
@thomasgodwell-optimizely ah sorry for letting this drag on. Yes, happy to rebase master and make those changes sometime this week.

rloomba:
@thomasgodwell-optimizely tooka pass at this refactor, let me know if you have any feedback.

oakbani:
@msohailhussain  and @thomaszurkan-optimizely  Tested FSC Suite here https://github.com/optimizely/ruby-sdk/pull/261

.travis.yml
```diff
@@ -43,7 +43,8 @@ jobs:
       install:
         - npm i -g markdown-spellcheck
       before_script:
-        - wget --quiet https://raw.githubusercontent.com/optimizely/mdspell-config/master/.spelling
+        # todo: change branch to master once merged.
+        - wget --quiet https://raw.githubusercontent.com/optimizely/mdspell-config/oakbani/ignore-for-ruby/.spelling
```
thomaszurkan-optimizely, line 47: can we merge and fix this?

thomaszurkan-optimizely:
This was move to PR #263 so that we could run FSC tests.  @rloomba thank you!  I made sure to include your credit in the PR and will also make another note to include you when I merge it.  Thanks again!

PR #250, opened 2020-05-05
fix: errors being raised without required args

oakbani:
## Summary
Some of the errors being throw without the required msg argument to initialize that error type. Expecting handle_error method inside unit tests doesn't initialize the error instance. This PR modifies the raise error assertion inside unit tests and fixes where required. 
Also removes an unused error type. 

## Test plan
All tests continue to pass.

### DIFF
--- a/spec/config_manager/http_project_config_manager_spec.rb
+++ b/spec/config_manager/http_project_config_manager_spec.rb
@@ -24,7 +24,7 @@ require 'optimizely/helpers/validator'
 require 'optimizely/logger'
 describe Optimizely::HTTPProjectConfigManager do
   let(:config_body_JSON) { OptimizelySpec::VALID_CONFIG_BODY_JSON }
-  let(:error_handler) { Optimizely::NoOpErrorHandler.new }
+  let(:error_handler) { Optimizely::RaiseErrorHandler.new }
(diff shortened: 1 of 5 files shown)


spec/config_manager/http_project_config_manager_spec.rb
```diff
@@ -24,7 +24,7 @@
 require 'optimizely/logger'
 describe Optimizely::HTTPProjectConfigManager do
   let(:config_body_JSON) { OptimizelySpec::VALID_CONFIG_BODY_JSON }
-  let(:error_handler) { Optimizely::NoOpErrorHandler.new }
+  let(:error_handler) { Optimizely::RaiseErrorHandler.new }
```
thomaszurkan-optimizely, line 27: What are the implications of changing the default behavior from no op error handler to one that raises an error?  Couldn't this update catch people off guard?

oakbani: This doesn't change the default Error Handler for the cient, if the client doesn't pass a custom error handler, still a NoOpErrorHandler will be used.  We are only using RaiseErrorHandler in unit tests so that we catch up any error that we throw without passing required params.

PR #255, opened 2020-06-03
feat: Added support for Authenticated Datafiles

zashraf1985:
## Summary
Added support for Authenticated Datafiles

## Test plan
Added new unit tests

### DIFF
--- a/spec/config_manager/http_project_config_manager_spec.rb
+++ b/spec/config_manager/http_project_config_manager_spec.rb
@@ -16,12 +16,7 @@
 #    limitations under the License.
 #
 require 'spec_helper'
-require 'optimizely/config_manager/http_project_config_manager'
-require 'optimizely/error_handler'
-require 'optimizely/exceptions'
-require 'optimizely/helpers/constants'
-require 'optimizely/helpers/validator'
-require 'optimizely/logger'
+
 describe Optimizely::HTTPProjectConfigManager do
   let(:config_body_JSON) { OptimizelySpec::VALID_CONFIG_BODY_JSON }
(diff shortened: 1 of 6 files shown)


zashraf1985:
Hey @thomaszurkan-optimizely . Thanks for reviewing quickly. I am already writing tests. That's what I mentioned in the PR summary too.

spec/config_manager/http_project_config_manager_spec.rb
```diff
@@ -468,4 +463,56 @@
...
+
+    it 'should add authorization header when auth token is provided' do
+      allow(Optimizely::Helpers::HttpUtils).to receive(:make_request)
+      @http_project_config_manager = Optimizely::HTTPProjectConfigManager.new(
+        sdk_key: 'valid_sdk_key',
+        datafile_access_token: 'the-token'
+      )
+      sleep 0.1
+      expect(Optimizely::Helpers::HttpUtils).to have_received(:make_request).with(anything, anything, anything, hash_including('Authorization' => 'Bearer the-token'), anything)
+    end
```
pthompson127, line 486: Just out of curiosity, is there a reason why this part of the spec is duplicated? Seems like the two blocks are exactly the same. 
Wondering if this is instead supposed to test the case where an invalid SDK key is provided (not sure if that is handled / error-checked elsewhere)

zashraf1985: Hey @pthompson127 ... good eye... its just a plain stupid mistake :). i wil fix it.

pthompson127: @zashraf1985 ahh gotcha :)

PR #259, opened 2020-07-06
fix(logs): Fixing log messages

oakbani:
## Summary
- This PR modifies log messages to be explicit when it's evaluating an experiment and when it's doing so for a Rollout. 

## Test plan
- Added additional unit test to assert audience logs for rollouts. 
- Existing tests should continue to pass.

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -810,7 +810,7 @@ module Optimizely
           end
         else
           @logger.log(Logger::DEBUG,
-                      "Feature '#{feature_flag_key}' for variation '#{variation['key']}' is not enabled. Returning the default variable value '#{variable_value}'.")
+                      "Feature '#{feature_flag_key}' is not enabled for user '#{user_id}'. Returning the default variable value '#{variable_value}'.")
         end
       else
         @logger.log(Logger::INFO,
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -806,7 +806,8 @@ module Optimizely
                         "Got variable value '#{variable_value}' for variable '#{variable['key']}' of feature flag '#{feature_flag_key}'.")
           else
             @logger.log(Logger::DEBUG,
-                        "Variable '#{variable['key']}' is not used in variation '#{variation['key']}'. Returning the default variable value '#{variable_value}'.")
+                        "Variable value is not defined. Returning the default variable value '#{variable_value}' for variable '#{variable['key']}'.")
+
           end
         else
           @logger.log(Logger::DEBUG,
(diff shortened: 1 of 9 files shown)


lib/optimizely.rb
```diff
@@ -810,7 +810,7 @@ def get_feature_variable_for_variation(feature_flag_key, feature_enabled, variat
           end
         else
           @logger.log(Logger::DEBUG,
-                      "Feature '#{feature_flag_key}' for variation '#{variation['key']}' is not enabled. Returning the default variable value '#{variable_value}'.")
+                      "Feature '#{feature_flag_key}' is not enabled for user '#{user_id}'. Returning the default variable value '#{variable_value}'.")
```
aliabbasrizvi, line 813: The changes are incomplete here. 

On line 809 it should say `Variable value is not defined. Returning the default variable value <variable value> for  variable <variable key>`

On line 916 it should say `User <user ID> was not bucketed into experiment or rollout for feature flag <feature flag key>. Returning the default variable value <variable value>`

lib/optimizely/audience.rb
```diff
@@ -24,23 +24,32 @@ module Optimizely
   module Audience
     module_function
 
-    def user_in_experiment?(config, experiment, attributes, logger)
-      # Determine for given experiment if user satisfies the audiences for the experiment.
+    def user_meets_audience_conditions?(config, experiment, attributes, logger, logging_hash = nil, logging_key = nil)
```
aliabbasrizvi, line 27: nit. `does_user_meet_audience_conditions`

lib/optimizely/audience.rb
```diff
@@ -24,23 +24,32 @@ module Optimizely
...
+      # Determine for given experiment/rollout rule if user satisfies the audiences.
```
aliabbasrizvi, line 28: nit. `satisfies the audience conditions`.

lib/optimizely/audience.rb
```diff
@@ -24,23 +24,32 @@ module Optimizely
...
       #
       # config - Representation of the Optimizely project config.
-      # experiment - Experiment for which visitor is to be bucketed.
+      # experiment - Experiment/Rollout rule for which visitor is to be bucketed.
```
aliabbasrizvi, line 31: nit. `...in which user is to be bucketed`

lib/optimizely/audience.rb
oakbani: I can rename it to match with other SDKs but the idiomatic code for ruby is to name methods like these with a question mark. See for example this link with for is_ and has_. https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Naming/PredicateName

PR #260, opened 2020-07-06
refactor: Added more logs to datafile manager

zashraf1985:
## Summary
Added more debug logs to datafile manager to make debugging easy.

## Test plan
All Existing unit and FSC tests should pass

### DIFF
--- a/lib/optimizely/config_manager/http_project_config_manager.rb
+++ b/lib/optimizely/config_manager/http_project_config_manager.rb
@@ -148,16 +148,18 @@ module Optimizely
     end
 
     def request_config
-      @logger.log(
-        Logger::DEBUG,
-        "Fetching datafile from #{@datafile_url}"
-      )
-      begin
-        headers = {}
-        headers['Content-Type'] = 'application/json'
-        headers['If-Modified-Since'] = @last_modified if @last_modified
-        headers['Authorization'] = "Bearer #{@access_token}" unless @access_token.nil?
+      @logger.log(Logger::DEBUG, "Fetching datafile from #{@datafile_url}")
+      headers = {}
+      headers['Content-Type'] = 'application/json'
+      headers['If-Modified-Since'] = @last_modified if @last_modified
+      headers['Authorization'] = "Bearer #{@access_token}" unless @access_token.nil?
 
+      # Cleaning headers before logging to avoid exposing authorization token
+      cleansed_headers = {}
+      headers.each { |key, value| cleansed_headers[key] = key == 'Authorization' ? '********' : value }
+      @logger.log(Logger::DEBUG, "Datafile request headers: #{cleansed_headers}")
+
+      begin
         response = Helpers::HttpUtils.make_request(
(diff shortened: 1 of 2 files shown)


lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -148,16 +148,18 @@ def fetch_datafile_config
...
-      begin
-        headers = {}
-        headers['Content-Type'] = 'application/json'
-        headers['If-Modified-Since'] = @last_modified if @last_modified
-        headers['Authorization'] = "Bearer #{@access_token}" unless @access_token.nil?
+      @logger.log(Logger::DEBUG, "Fetching datafile from #{@datafile_url}")
+      headers = {}
+      headers['Content-Type'] = 'application/json'
+      headers['If-Modified-Since'] = @last_modified if @last_modified
+      headers['Authorization'] = "Bearer #{@access_token}" unless @access_token.nil?
```
thomaszurkan-optimizely, line 155: We probably shouldn't output auth tokens to the logs.

lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -178,9 +183,14 @@ def request_config
         return
       end
 
-      @last_modified = response[Helpers::Constants::HTTP_HEADERS['LAST_MODIFIED']]
-
-      config = DatafileProjectConfig.create(response.body, @logger, @error_handler, @skip_json_validation) if response.body
+      if response_code >= 200 && response_code < 400
+        @logger.log(Logger::DEBUG, 'Successfully fetched datafile, generating Project config')
+        config = DatafileProjectConfig.create(response.body, @logger, @error_handler, @skip_json_validation)
+        @last_modified = response[Helpers::Constants::HTTP_HEADERS['LAST_MODIFIED']]
```
thomaszurkan-optimizely, line 189: what happens if last modified is nil?

lib/optimizely/config_manager/http_project_config_manager.rb
zashraf1985: You are right, that is exactly what i did. This is why i cleansed the headers before logging and replaced the value of `Authorization` header to `***` right below this line.

lib/optimizely/config_manager/http_project_config_manager.rb
zashraf1985: It is assigned as nil. I just moved this line of code inside the check. assuming it was working correctly earlier as well without a `nil` check

lib/optimizely/config_manager/http_project_config_manager.rb
thomaszurkan-optimizely: ha, i saw that. i deleted my comment.  but, it would be nice to add a simple logger spy test to see that was outputted :)

zashraf1985: yup.. good idea.. doing that now.

zashraf1985: Added tests

PR #269, opened 2020-09-04
fix: check pre-release and build versions

oakbani:
## Summary
This PR adds some additional checks when  figuring out if the given version is a build or a prerelease. 
For a pre-release version . '-' should come before '+' ( if plus exists). 
For a build version: '+' should come before '-' (if minus exists)

## Test plan
Added unit tests

## Issues
- "THING-1234" or "Fixes #123"

### DIFF
--- a/lib/optimizely/semantic_version.rb
+++ b/lib/optimizely/semantic_version.rb
@@ -27,14 +27,43 @@ module Optimizely
     module_function
 
     def pre_release?(target)
-      # Method to check if the given version contains "-"
+      # Method to check if the given version is a prerelease
       #
       # target - String representing semantic version
       #
-      # Returns true if the given version does contain "-"
+      # Returns true if the given version is a prerelease
       #         false if it doesn't
 
-      target.include? SEMVER_PRE_RELEASE
+      raise unless target.is_a? String
+
+      prerelease_index = target.index(SEMVER_PRE_RELEASE)
+      build_index = target.index(SEMVER_BUILD)
+
+      return false if prerelease_index.nil?
+      return true if build_index.nil?
+
+      # when both operators are present prerelease should precede the build operator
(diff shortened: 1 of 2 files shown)


lib/optimizely/semantic_version.rb
```diff
@@ -27,14 +27,43 @@ module SemanticVersion
...
+      raise unless target.is_a? String
+
+      prerelease_index = target.index(SEMVER_PRE_RELEASE)
+      build_index = target.index(SEMVER_BUILD)
+
+      return false if build_index.nil?
+      return true if prerelease_index.nil?
+
+      # when both operators are present build should precede the prerelease operator
+      build_index < prerelease_index
```
aliabbasrizvi, line 66: Have we considered just having a single method called `prerelease_or_build` and that just determines if version is prerelease or build vs doing similar things twice.

aliabbasrizvi: Calling the index method is not cheap and we can technically halve the number of times we call it.

thomaszurkan-optimizely: @aliabbasrizvi I would generally agree with you.  However, you need to keep in context your context :D.  Walking a small string like a version string (finding first index is going to be quite small (e.g. 2.0.0 or 2.0.0- or 2.0.0+).  We could always just check the 6th element.  But again, in general you are walking a tiny string.

thomaszurkan-optimizely (approved):
LGTM

lib/optimizely/semantic_version.rb
aliabbasrizvi: Sure. I accept that it is small. Ok in current form.

aliabbasrizvi (approved):

PR #275, opened 2020-11-10
feat: added "enabled" field to decision metadata structure

zashraf1985:
## Summary
1. Added "enabled" field to decision metadata structure
2. Setting "enabled" to true for activate() since we check for non nil variation. 

## Test Plan
1. Manuall tested thoroughly
2. Fixed existing unit tests

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -142,7 +142,7 @@ module Optimizely
       experiment = config.get_experiment_from_key(experiment_key)
       send_impression(
         config, experiment, variation_key, '', experiment_key,
-        Optimizely::DecisionService::DECISION_SOURCES['EXPERIMENT'], user_id, true, attributes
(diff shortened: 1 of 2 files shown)


Mat001 (commented):
LGTM

lib/optimizely.rb
```diff
@@ -142,7 +142,7 @@ def activate(experiment_key, user_id, attributes = nil)
       experiment = config.get_experiment_from_key(experiment_key)
       send_impression(
         config, experiment, variation_key, '', experiment_key,
-        Optimizely::DecisionService::DECISION_SOURCES['EXPERIMENT'], user_id, attributes
+        Optimizely::DecisionService::DECISION_SOURCES['EXPERIMENT'], user_id, true, attributes
```
pawels-optimizely, line 145: can we add this parameter after experiment key ? I think adding that between user_id and attributes is a wrong place

pawels-optimizely (approved):

PR #276, opened 2020-11-16
(fix): allow user_attributes to be evaluated with symbol or string keys

BuddyLReno:
## Summary
- Change the default_proc of the user_attributes hash to check for string keys or symbol keys and return the appropriate value.

The SDK allows a user to send the user_attributes hash with symbol keys or with string keys. If passing a hash with symbol keys, the evaluate check will always return nil as the method only checks for string keys. This will enable the attribute evaluation to evaluate string or symbol keys.

Specifically, this check here makes sure the hash has symbol or string keys: https://github.com/optimizely/ruby-sdk/blob/73fdcf663396384f1307a9a8cc3e75d876342b97/lib/optimizely/helpers/validator.rb#L37-L50 but it never evaluates symbol keys.

## Test plan
- Added unit tests to pass in user_attributes with a symbol hash and ensure new tests and existing tests still passed.

## Issues
- none opened at this time.

### DIFF
--- a/lib/optimizely/custom_attribute_condition_evaluator.rb
+++ b/lib/optimizely/custom_attribute_condition_evaluator.rb
@@ -58,6 +58,18 @@ module Optimizely
     def initialize(user_attributes, logger)
       @user_attributes = user_attributes
       @logger = logger
+
+      # configure user_attributes to access with string or symbol
+      @user_attributes.default_proc = proc do |h, k|
+        case k
+        when String
+          sym = k.to_sym
+          h[sym] if h.key?(sym)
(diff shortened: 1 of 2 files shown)


BuddyLReno:
Would love to know if there's a better way to handle this! Just took a stab to see if I could enable that functionality. :)

BuddyLReno:
Realized I created the PR incorrectly. Closing to reopen with new pr. Sorry about that!

BuddyLReno:
Reopened as #277 with the proper branch name on my fork.

PR #281, opened 2021-06-08
Feat: added SDKkey and environmentKey in datafile

ozayr-zaviar:
## Summary
- Sdk and environment keys are going to be parse in sdk and will be sent back in optimizelyConfig

## Test plan
All unit tests and FSC tests should pass

### DIFF
--- a/lib/optimizely/config/datafile_project_config.rb
+++ b/lib/optimizely/config/datafile_project_config.rb
@@ -83,6 +83,8 @@ module Optimizely
       @anonymize_ip = config.key?('anonymizeIP') ? config['anonymizeIP'] : false
       @bot_filtering = config['botFiltering']
       @revision = config['revision']
+      @sdk_key = config,fetch('sdkKey', nil)
+      @environment_key = config.fetch('environmentKey', nil)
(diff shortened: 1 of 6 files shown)


lib/optimizely/config/datafile_project_config.rb
```diff
@@ -38,6 +38,8 @@ class DatafileProjectConfig < ProjectConfig
     attr_reader :anonymize_ip
     attr_reader :bot_filtering
     attr_reader :revision
+    attr_reader :sdk_key
```
msohailhussain, line 41: update header.

lib/optimizely/config/datafile_project_config.rb
```diff
@@ -194,6 +198,26 @@ def experiment_running?(experiment)
       RUNNING_EXPERIMENT_STATUS.include?(experiment['status'])
     end
 
+    def get_sdk_key()
```
zashraf1985, line 201: why create these getters when you have already declared read only attributes and using those directly?

zashraf1985 (changes requested):

msohailhussain (approved):
lgtm

PR #282, opened 2021-07-12
feat: Duplicate experiment key issue with multi feature flag

ozayr-zaviar:
## Summary
- Fixed an issue related to duplicate experiment_key.
- While trying to get variation from the variation_key_map, it was unable to find because the latest experimentKey was overriding the previous one.

## TestPlan
- Run with FSC 
https://travis-ci.com/github/optimizely/fullstack-sdk-compatibility-suite/jobs/523753247

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 #
-#    Copyright 2016-2020, Optimizely and contributors
+#    Copyright 2016-2021, Optimizely and contributors
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
@@ -877,12 +877,14 @@ module Optimizely
       experiment = config.get_experiment_from_key(experiment_key)
       return nil if experiment.nil?
 
+      experiment_id = experiment['id']
+
       return nil unless user_inputs_valid?(attributes)
 
-      variation_id, = @decision_service.get_variation(config, experiment_key, user_id, attributes)
+      variation_id, = @decision_service.get_variation(config, experiment_id, user_id, attributes)
       variation = config.get_variation_from_id(experiment_key, variation_id) unless variation_id.nil?
       variation_key = variation['key'] if variation
-      decision_notification_type = if config.feature_experiment?(experiment['id'])
+      decision_notification_type = if config.feature_experiment?(experiment_id)
                                      Helpers::Constants::DECISION_NOTIFICATION_TYPES['FEATURE_TEST']
                                    else
                                      Helpers::Constants::DECISION_NOTIFICATION_TYPES['AB_TEST']
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -1081,6 +1081,7 @@ module Optimizely
       end
 
       experiment_id = experiment['id']
+      experiment_key = experiment['key']
 
       variation_id = ''
       variation_id = config.get_variation_id_from_key_by_experiment_id(experiment_id, variation_key) if experiment_id != ''
@@ -1097,7 +1098,7 @@ module Optimizely
       @event_processor.process(user_event)
       return unless @notification_center.notification_count(NotificationCenter::NOTIFICATION_TYPES[:ACTIVATE]).positive?
 
-      @logger.log(Logger::INFO, "Activating user '#{user_id}' in experiment '#{experiment_id}'.")
+      @logger.log(Logger::INFO, "Activating user '#{user_id}' in experiment '#{experiment_key}'.")
 
       experiment = nil if experiment_id == ''
       variation = nil
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -1078,10 +1078,10 @@ module Optimizely
         }
       end
 
-      experiment_key = experiment['key']
+      experiment_id = experiment['id']
 
       variation_id = ''
-      variation_id = config.get_variation_id_from_key(experiment_key, variation_key) if experiment_key != ''
+      variation_id = config.get_variation_id_from_key_by_experiment_id(experiment_id, variation_key) if experiment_id != ''
 
       metadata = {
         flag_key: flag_key,
@@ -1095,11 +1095,11 @@ module Optimizely
       @event_processor.process(user_event)
       return unless @notification_center.notification_count(NotificationCenter::NOTIFICATION_TYPES[:ACTIVATE]).positive?
 
-      @logger.log(Logger::INFO, "Activating user '#{user_id}' in experiment '#{experiment_key}'.")
+      @logger.log(Logger::INFO, "Activating user '#{user_id}' in experiment '#{experiment_id}'.")
 
-      experiment = nil if experiment_key == ''
+      experiment = nil if experiment_id == ''
       variation = nil
-      variation = config.get_variation_from_id(experiment_key, variation_id) unless experiment.nil?
+      variation = config.get_variation_from_id_by_experiment_id(experiment_id, variation_id) unless experiment.nil?
       log_event = EventFactory.create_log_event(user_event, @logger)
       @notification_center.send_notifications(
         NotificationCenter::NOTIFICATION_TYPES[:ACTIVATE],
--- a/lib/optimizely/config/datafile_project_config.rb
+++ b/lib/optimizely/config/datafile_project_config.rb
@@ -113,7 +113,7 @@ module Optimizely
           @experiments.push(exp.merge('groupId' => key))
         end
       end
-      @experiment_key_map = generate_key_map(@experiments, 'key')
+      @experiment_key_map = {}
       @experiment_id_map = generate_key_map(@experiments, 'id')
       @audience_id_map = generate_key_map(@audiences, 'id')
       @audience_id_map = @audience_id_map.merge(generate_key_map(@typed_audiences, 'id')) unless @typed_audiences.empty?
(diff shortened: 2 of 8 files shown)


msohailhussain (commented):
please review the logic, you can extract experiment_key_map from the experiment_id_map, other way will cause issues.

mnoman09 (changes requested):
Overall LGTM, Just update headers.

jaeopt (changes requested):
This fixes the delivery rule-key conflict. I still see potential issue with experiment key conflicts. I believe they'll covered by new FSC test cases.

mnoman09 (changes requested):

lib/optimizely.rb
```diff
@@ -1095,11 +1097,11 @@ def send_impression(config, experiment, variation_key, flag_key, rule_key, enabl
       @event_processor.process(user_event)
       return unless @notification_center.notification_count(NotificationCenter::NOTIFICATION_TYPES[:ACTIVATE]).positive?
 
-      @logger.log(Logger::INFO, "Activating user '#{user_id}' in experiment '#{experiment_key}'.")
+      @logger.log(Logger::INFO, "Activating user '#{user_id}' in experiment '#{experiment_id}'.")
```
jaeopt, line 1100: keep experiment_key in log

lib/optimizely/decision_service.rb
```diff
@@ -68,24 +68,23 @@ def get_variation(project_config, experiment_key, user_id, attributes = nil, dec
...
       decide_reasons.push(*bucketing_id_reasons)
       # Check to make sure experiment is active
-      experiment = project_config.get_experiment_from_key(experiment_key)
+      experiment = project_config.get_experiment_from_id(experiment_id)
       return nil, decide_reasons if experiment.nil?
 
-      experiment_id = experiment['id']
       unless project_config.experiment_running?(experiment)
-        message = "Experiment '#{experiment_key}' is not running."
+        message = "Experiment '#{experiment_id}' is not running."
```
jaeopt, line 75: keep experiment_key for log and message

lib/optimizely/decision_service.rb
```diff
@@ -108,7 +107,7 @@ def get_variation(project_config, experiment_key, user_id, attributes = nil, dec
       user_meets_audience_conditions, reasons_received = Audience.user_meets_audience_conditions?(project_config, experiment, attributes, @logger)
       decide_reasons.push(*reasons_received)
       unless user_meets_audience_conditions
-        message = "User '#{user_id}' does not meet the conditions to be in experiment '#{experiment_key}'."
+        message = "User '#{user_id}' does not meet the conditions to be in experiment '#{experiment_id}'."
```
jaeopt, line 110: keep experiment_key in all messages here and others

jaeopt (changes requested):
Looks good to change experiment-key to -id based. 
For log and decision messages, please keep all 'experiment-key'as is (instead of using 'experiment-id' which is not helpful to users).

lib/optimizely.rb
zashraf1985: I agree with @jaeopt on this. Also adding experiment_id can actually help too

lib/optimizely/config/datafile_project_config.rb
```diff
@@ -145,8 +149,10 @@ def initialize(datafile, logger, error_handler)
 
           @variation_id_to_variable_usage_map[variation_id] = generate_key_map(variation_variables, 'id')
         end
-        @variation_id_map[key] = generate_key_map(variations, 'id')
-        @variation_key_map[key] = generate_key_map(variations, 'key')
+        @variation_id_map[exp['key']] = generate_key_map(variations, 'id')
+        @variation_key_map[exp['key']] = generate_key_map(variations, 'key')
+        @variation_id_map_by_experiment_id[id] = generate_key_map(variations, 'id')
+        @variation_key_map_by_experiment_id[id] = generate_key_map(variations, 'key')
```
zashraf1985, line 155: These variable names are really confusing. Looking at `variation_id_map`, it looks like its a map of variations with variation ids as keys but looks like the key is the experiment id which contains all the variations for that experiment. We might want to consider refactoring it later for better and more intuitive naming. Or if possible and if it makes sense to you , can you rename these variables to reflect more of what they actually are. `variation_id_map`  can probably be `variation_map_by_experiment_id` etc. You can certainly think of a better variable name or can also leave it like that to refactor it later, i wont block the review on this.

lib/optimizely/config/datafile_project_config.rb
```diff
@@ -281,6 +302,52 @@ def get_variation_from_id(experiment_key, variation_id)
       nil
     end
 
+    def get_variation_from_id_by_experiment_id(experiment_id, variation_id)
```
zashraf1985, line 305: Can variation ids duplicate across experiments? why do we need `experiment_id` if we have `variation_id`. shouldn't `variation_id` be enought to uniquely identify and fetch a variation?

lib/optimizely/config/datafile_project_config.rb
```diff
@@ -281,6 +302,52 @@ def get_variation_from_id(experiment_key, variation_id)
...
+        @error_handler.handle_error InvalidVariationError
+        return nil
+      end
+
+      @logger.log Logger::ERROR, "Experiment id '#{experiment_id}' is not in datafile."
+      @error_handler.handle_error InvalidExperimentError
+      nil
+    end
+
+    def get_variation_id_from_key_by_experiment_id(experiment_id, variation_key)
```
zashraf1985, line 328: This makes sense because `variation_key` can probably duplicate across experiments.

lib/optimizely/config/datafile_project_config.rb
```diff
@@ -304,17 +371,17 @@ def get_variation_id_from_key(experiment_key, variation_key)
...
       # experiment_key - String Key representing the experiment
       #
       # Returns whitelisted variations for the experiment or nil
 
-      experiment = @experiment_key_map[experiment_key]
+      experiment = @experiment_id_map[experiment_id]
       return experiment['forcedVariations'] if experiment
 
-      @logger.log Logger::ERROR, "Experiment key '#{experiment_key}' is not in datafile."
+      @logger.log Logger::ERROR, "Experiment key '#{experiment_id}' is not in datafile."
```
zashraf1985, line 384: It will be helpful to keep both `experiment_id` and `experiment_key` in the log

lib/optimizely/decision_service.rb
zashraf1985: Or Keep both.

lib/optimizely/decision_service.rb
zashraf1985: Yes, Agree! An i think you should keep both id and key in the messages

lib/optimizely/decision_service.rb
```diff
@@ -186,13 +185,13 @@ def get_variation_for_feature_experiment(project_config, feature_flag, user_id,
...
-        experiment_key = experiment['key']
-        variation_id, reasons_received = get_variation(project_config, experiment_key, user_id, attributes, decide_options)
+        experiment_id = experiment['id']
+        variation_id, reasons_received = get_variation(project_config, experiment_id, user_id, attributes, decide_options)
         decide_reasons.push(*reasons_received)
 
         next unless variation_id
 
-        variation = project_config.variation_id_map[experiment_key][variation_id]
+        variation = project_config.variation_id_map_by_experiment_id[experiment_id][variation_id]
```
zashraf1985, line 194: Why accessing it directly from the map when you have an accessor method?

lib/optimizely/config/datafile_project_config.rb
```diff
@@ -145,8 +149,10 @@ def initialize(datafile, logger, error_handler)
 
           @variation_id_to_variable_usage_map[variation_id] = generate_key_map(variation_variables, 'id')
         end
-        @variation_id_map[key] = generate_key_map(variations, 'id')
-        @variation_key_map[key] = generate_key_map(variations, 'key')
+        @variation_id_map[exp['key']] = generate_key_map(variations, 'id')
+        @variation_key_map[exp['key']] = generate_key_map(variations, 'key')
+        @variation_id_map_by_experiment_id[id] = generate_key_map(variations, 'id')
```
zashraf1985, line 154: why do we need it by experiment id? i assume variation ids are unique? if yes, then a simple map with variation ids as keys and variations as values would suffice? It was probably required in case of variation keys because keys can be reused across the experiments and you needed to specify which experiment to look into, but for variation id, i am not sure if experiment info is needed at all?

zashraf1985 (changes requested):

lib/optimizely.rb
jaeopt: Having both will be helpful, but for consistency with all other SDKs, let's stick to "key" only in logs and messages

jaeopt (changes requested):
I do not see the last reviews addressed yet. Added some more comments.

lib/optimizely/config/datafile_project_config.rb
msohailhussain: let's talk offline.

lib/optimizely/config/datafile_project_config.rb
ozayr-zaviar: Cant use experiment key here as we are no longer sending key in `get_whitelisted_variations` instead we are sending the ID. If the given ID is wrong we cannot get an experiment and hence we can not get the key.

lib/optimizely/config/datafile_project_config.rb
ozayr-zaviar: Yes as far as I know variation Ids are unique but I was following this method [`def get_variation_from_id(experiment_key, variation_id)`](https://github.com/optimizely/ruby-sdk/blob/697e13d9a67c2f4ddde3764898bbd68b1d9b52af/lib/optimizely/config/datafile_project_config.rb#L282)

jaeopt (approved):
LGTM

zashraf1985 (approved):

msohailhussain (approved):
lgtm

PR #289, opened 2021-11-16
fix: optimizelyClient validity check fixed

ozayr-zaviar:
## Summary
- `get_optimizely_config` replaced with `is_valid` in forced decision `optimizelyClient` object check.

## Test plan
- All unit tests and FSC tests should pass.


jaeopt (approved):
LGTM

PR #292, opened 2021-12-14
fix: Forced variation not available in experiment

ozayr-zaviar:
## Summary
- Crash fixes and order set for variation by flag first then experiment

## Test plan
- All unit tests and FSC should pass

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -298,6 +298,15 @@ module Optimizely
       decisions
     end
 
+    # Gets variation using variation key or id and flag key.
+    #
+    # @param flag_key - flag key from which the variation is required.
+    # @param target_value - variation value either id or key that will be matched.
+    # @param attribute - string representing variation attribute.
+    #
+    # @return [variation]
+    # @return [nil] if no variation found in flag_variation_map.
+
     def get_flag_variation(flag_key, target_value, attribute)
       project_config.get_variation_from_flag(flag_key, target_value, attribute)
     end
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -1107,13 +1107,12 @@ module Optimizely
       experiment_id = experiment['id']
       experiment_key = experiment['key']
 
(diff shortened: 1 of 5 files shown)


Mat001 (approved):
LGTM

lib/optimizely.rb
```diff
@@ -1098,12 +1098,13 @@ def send_impression(config, experiment, variation_key, flag_key, rule_key, enabl
...
-        varaition = get_flag_variation_by_key(flag_key, variation_key)
-        variation_id = varaition ? varaition['id'] : ''
-      end
+      variation = get_flag_variation(flag_key, variation_key, 'key')
+
+      variation_id = if !variation
+                       experiment_id != '' ? config.get_variation_id_from_key_by_experiment_id(experiment_id, variation_key) : ''
+                     else
+                       variation ? variation['id'] : ''
+                     end
```
jaeopt, line 1107:
```suggestion
      variation_id = variation ? variation['id'] : ''
```
If it's null from get_flag_variagtion(), then get_variation_id_from_key_by_experiment_id() will also return null.

jaeopt (changes requested):
A fix suggested. Also need a unit test covering the case.

lib/optimizely.rb
```diff
@@ -298,8 +298,8 @@ def decide_for_keys(user_context, keys, decide_options = [])
       decisions
     end
 
-    def get_flag_variation_by_key(flag_key, variation_key)
-      project_config.get_variation_from_flag(flag_key, variation_key)
+    def get_flag_variation(flag_key, target_value, attribute)
```
msohailhussain, line 301: please add description of this method.

lib/optimizely.rb
msohailhussain: This is independent of `flag_key`, for `a/b` experiments this will return non-null and `get_flag_variation` will return null.

jaeopt (changes requested):
The unit test looks good! A suggestion for refactoring.

jaeopt (approved):
LGTM

PR #293, opened 2021-12-22
refactor: moved validated forced decision to decision service.

ozayr-zaviar:
Summary
-------
- `find_validated_forced_decision` moved to decision service.

Test plan
---------
- All unit tests and FSC should pass.

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 #
-#    Copyright 2016-2021, Optimizely and contributors
+#    Copyright 2016-2022, Optimizely and contributors
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -199,7 +199,7 @@ module Optimizely
       experiment = nil
       decision_source = Optimizely::DecisionService::DECISION_SOURCES['ROLLOUT']
       context = Optimizely::OptimizelyUserContext::OptimizelyDecisionContext.new(key, nil)
-      variation, reasons_received = @decision_service.validated_forced_decision(context, user_context)
+      variation, reasons_received = @decision_service.validated_forced_decision(config, context, user_context)
       reasons.push(*reasons_received)
 
       if variation
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -199,7 +199,7 @@ module Optimizely
       experiment = nil
       decision_source = Optimizely::DecisionService::DECISION_SOURCES['ROLLOUT']
       context = Optimizely::OptimizelyUserContext::OptimizelyDecisionContext.new(key, nil)
-      variation, reasons_received = user_context.find_validated_forced_decision(context)
+      variation, reasons_received = @decision_service.validated_forced_decision(context, user_context)
       reasons.push(*reasons_received)
 
       if variation
--- a/lib/optimizely/decision_service.rb
+++ b/lib/optimizely/decision_service.rb
@@ -418,7 +418,7 @@ module Optimizely
     end
 
     def validated_forced_decision(context, user_context)
-      decision = user_context.find_forced_decision(context)
+      decision = user_context.get_forced_decision(context)
       flag_key = context[:flag_key]
       rule_key = context[:rule_key]
       variation_key = decision ? decision[:variation_key] : decision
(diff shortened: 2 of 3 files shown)


lib/optimizely.rb
```diff
@@ -199,7 +199,7 @@ def decide(user_context, key, decide_options = [])
       experiment = nil
       decision_source = Optimizely::DecisionService::DECISION_SOURCES['ROLLOUT']
       context = Optimizely::OptimizelyUserContext::OptimizelyDecisionContext.new(key, nil)
-      variation, reasons_received = user_context.find_validated_forced_decision(context)
+      variation, reasons_received = @decision_service.validated_forced_decision(context, user_context)
```
yasirfolio3, line 202: please keep the method name `find_validated_forced_decision`

lib/optimizely/decision_service.rb
```diff
@@ -417,6 +417,28 @@ def get_forced_variation(project_config, experiment_key, user_id)
       [variation, decide_reasons]
     end
 
+    def validated_forced_decision(context, user_context)
+      decision = user_context.find_forced_decision(context)
+      flag_key = context[:flag_key]
+      rule_key = context[:rule_key]
```
yasirfolio3, line 423: can we move this code inside the if block since it is only used there?

lib/optimizely/decision_service.rb
```diff
@@ -417,6 +417,28 @@ def get_forced_variation(project_config, experiment_key, user_id)
...
+      variation_key = decision ? decision[:variation_key] : decision
+      reasons = []
+      target = rule_key ? "flag (#{flag_key}), rule (#{rule_key})" : "flag (#{flag_key})"
+      if variation_key
+        variation = user_context.optimizely_client.get_flag_variation(flag_key, variation_key, 'key')
+        if variation
+          reason = "Variation (#{variation_key}) is mapped to #{target} and user (#{user_context.user_id}) in the forced decision map."
+          reasons.push(reason)
+          return variation, reasons
```
yasirfolio3, line 432: this code can be moved to line 436, that way we won't have `reasons.push(reason)` twice.

lib/optimizely/optimizely_user_context.rb
```diff
@@ -26,6 +26,7 @@ class OptimizelyUserContext
     attr_reader :forced_decisions
     attr_reader :OptimizelyDecisionContext
     attr_reader :OptimizelyForcedDecision
+    attr_reader :optimizely_client
```
yasirfolio3, line 29: why is this needed when we are only removing a method here?

yasirfolio3 (changes requested):

lib/optimizely/optimizely_user_context.rb
yasirfolio3: Also update the Copyright header with `2020-2021`

yasirfolio3 (changes requested):

lib/optimizely/decision_service.rb
```diff
@@ -417,6 +417,28 @@ def get_forced_variation(project_config, experiment_key, user_id)
       [variation, decide_reasons]
     end
 
+    def validated_forced_decision(context, user_context)
+      decision = user_context.find_forced_decision(context)
```
jaeopt, line 421: Remove `find_forced_decision` and use `get_forced_decision`. Since we remove sanity checking, they're now identical.

jaeopt (changes requested):
A couple of changes suggested

lib/optimizely/decision_service.rb
```diff
@@ -417,6 +417,28 @@ def get_forced_variation(project_config, experiment_key, user_id)
...
 
+    def validated_forced_decision(context, user_context)
+      decision = user_context.get_forced_decision(context)
+      flag_key = context[:flag_key]
+      rule_key = context[:rule_key]
+      variation_key = decision ? decision[:variation_key] : decision
+      reasons = []
+      target = rule_key ? "flag (#{flag_key}), rule (#{rule_key})" : "flag (#{flag_key})"
+      if variation_key
+        variation = user_context.optimizely_client.get_flag_variation(flag_key, variation_key, 'key')
```
jaeopt, line 428: I don't what happened to my previous comments on this line. github bug? :)
We need to change this line to same as other SDKs - pass 'config' as a parameter and move `get_flag_variation` to `config`, so we can remove dependency to `user_context.optimizely_client` here.

jaeopt (changes requested):
See comments

jaeopt (approved):
LGTM

yasirfolio3 (approved):

PR #296, opened 2022-01-27
enhancement: Generate OptimizelyConfig object on API Call instead of SDK initialization

### LINKED ISSUE #295: Slowness after upgrading to 3.9
Hey there, we recently bumped to 3.9 and noticed latency spike (~double) along a critical code-path. We have json schema validation disabled here for performance reasons. 

As a comparison, here's a flamegraph I made where our code uses

Release version 3.8 of the ruby-sdk gem:
Screen Shot 2022-01-24 at 10 50 32 AM

vs. 3.9.
Screen Shot 2022-01-24 at 10 50 43 AM

I see a lot of cycles spent in `Optimizely::StaticProfileManager`, possibly tied to [this change](https://github.com/optimizely/ruby-sdk/commit/3a30a000aadbde5633cb183ff62593e4f0b2141a)

I was wondering if anyone's noticed something similar/ could give me more context on the change here and what we might be able to do to address the performance issues

zashraf1985:
## Summary
`OptimizelyConfig` object is generated when SDK is being initialized. This operation is heavy and blocks the SDK from initializing quickly. This object is only used when users explicitly access it using `get_optimizely_config` API call. This PR moves the creation of this object from SDK initialization to the actual API call.

## Test plan
- Manually tested thoroughly
- All unit tests pass
- All FSC tests pass

## Issues
fixes #295

### DIFF
--- a/lib/optimizely/config_manager/http_project_config_manager.rb
+++ b/lib/optimizely/config_manager/http_project_config_manager.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 #
-#    Copyright 2019-2020, Optimizely and contributors
+#    Copyright 2019-2020, 2022, Optimizely and contributors
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -33,7 +33,7 @@ module Optimizely
   class HTTPProjectConfigManager < ProjectConfigManager
     # Config manager that polls for the datafile and updated ProjectConfig based on an update interval.
 
-    attr_reader :stopped, :optimizely_config
+    attr_reader :stopped
```
msohailhussain, line 36: please update copyright :)

jaeopt (approved):
LGTM

PR #306, opened 2022-08-02
feat: add lru cache

andrewleap-optimizely:
Summary
-------

Add a generic LRU cache to support efficient caching of the ODP segments results. 

- Add Least Recently Used (LRU) cache
- design to allow developer to pass in different caching implementation. Provide default.

Test plan
---------
Added:
- lru_cache_spec.rb

Ticket:
---------
[OASIS-8435](https://optimizely.atlassian.net/browse/OASIS-8435)


lib/optimizely/odp/lru_cache.rb
```diff
@@ -0,0 +1,116 @@
...
+
+      @cache_mutex.synchronize do
+        return nil unless @map.include?(key)
+
+        old_element = @map[key]
+        @map.delete(key)
+        return nil if old_element.stale?(@timeout)
+
+        element = CacheElement.new(old_element.value)
+        @map[key] = element
```
jaeopt, line 52: It looks like ruby hash keeps insertion order for 1.9+. Min ruby version we support?

lib/optimizely/odp/lru_cache.rb
```diff
@@ -0,0 +1,116 @@
...
+      return nil if @capacity <= 0
+
+      @cache_mutex.synchronize do
+        return nil unless @map.include?(key)
+
+        old_element = @map[key]
+        @map.delete(key)
+        return nil if old_element.stale?(@timeout)
+
+        element = CacheElement.new(old_element.value)
```
jaeopt, line 51: Timestamp is not supposed to change for lookup. This new creation will overwrite the old timestamp.

jaeopt (changes requested):
Looks great. A fix suggested.

lib/optimizely/odp/lru_cache.rb
andrewleap-optimizely: That's one of the main reasons I pushed for updating the minimum supported ruby version. As of 4.0.0, we now only support >=2.7

jaeopt: @andrewleap-optimizely That min version info should be in README.md file and our docs. Updated?

andrewleap-optimizely: The minimum version is listed on the rubygems page but yea we should add it to the readme as well. Looks like a lot of the sdks are missing that information on their readme/docs, may be worth discussing.

lib/optimizely/odp/lru_cache.rb
andrewleap-optimizely: good catch. Fixed and added test to confirm.

jaeopt (approved):
LGTM

zashraf1985 (approved):
Looks Great!

PR #310, opened 2022-08-16
feat: add odp segment manager

andrewleap-optimizely:
## Summary
Adds OdpSegmentManager to provide internal services for fetching segments from the ODP server. 
- Provides an internal api for fetching segments with given { api_key, api_host, customer key, customer value, segments subset, and options }.
- Manages a LRU cache for fetched segments (size and timeout are configurable).
- On segment fetch request, looks up the cache for the user key first. On cache hit (or timeout not expired) it returns the cached segments. Otherwise, it sends a fetch request to the ODP server and caches the result.
- Supports options for cache lookup (IGNORE_CACHE, RESET_CACHE).

## Test plan
- Added  odp_segment_manager_spec.rb
## Ticket
[OASIS-8446](https://optimizely.atlassian.net/browse/OASIS-8446)

### DIFF
--- a/lib/optimizely/odp/odp_config.rb
+++ b/lib/optimizely/odp/odp_config.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+
+#
+#    Copyright 2022, Optimizely and contributors
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+require 'optimizely/logger'
+
+module Optimizely
+  class OdpConfig
+    # Contains configuration used for ODP integration.
+    #
+    # @param api_host - The host URL for the ODP audience segments API (optional). If not provided, SDK will use the default host in datafile.
+    # @param api_key - The public API key for the ODP account from which the audience segments will be fetched (optional). If not provided, SDK will use the default publicKey in datafile.
+    # @param segments_to_check - An array of all ODP segments used in the current datafile (associated with api_host/api_key).
+    # @param odp_enabled - A boolean value indicating whether odp is enabled for the project or not.
+    def initialize(api_key = nil, api_host = nil, segments_to_check = [])
+      @api_key = api_key
+      @api_host = api_host
+      @segments_to_check = segments_to_check
+      @odp_enabled = !api_key.nil? && !api_host.nil? ? true : false
+      @mutex = Mutex.new
+    end
+
+    # Replaces the existing configuration
+    #
+    # @param api_host - The host URL for the ODP audience segments API (optional). If not provided, SDK will use the default host in datafile.
+    # @param api_key - The public API key for the ODP account from which the audience segments will be fetched (optional). If not provided, SDK will use the default publicKey in datafile.
+    # @param segments_to_check - An array of all ODP segments used in the current datafile (associated with api_host/api_key).
+    #
+    # @return - True if the provided values were different than the existing values.
+
+    def update(api_key = nil, api_host = nil, segments_to_check = [])
+      @mutex.synchronize do
+        @odp_enabled = !api_key.nil? && !api_host.nil? ? true : false
+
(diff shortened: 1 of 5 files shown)


lib/optimizely/odp/odp_config.rb
```diff
@@ -0,0 +1,122 @@
...
+    #
+    # @param api_host - The host URL for the ODP audience segments API (optional). If not provided, SDK will use the default host in datafile.
+    # @param api_key - The public API key for the ODP account from which the audience segments will be fetched (optional). If not provided, SDK will use the default publicKey in datafile.
+    # @param segments_to_check - An array of all ODP segments used in the current datafile (associated with api_host/api_key).
+    #
+    # @return - True if the provided values were different than the existing values.
+
+    def update(api_key = nil, api_host = nil, segments_to_check = [])
+      @mutex.synchronize do
+        @odp_enabled = !api_key.nil? && !api_host.nil? ? true : false
```
zashraf1985, line 47: Might sound like being nitpicking but just curios how it will play out if someone later passes a null apiKey or host through the setter, should this be updated to a method that evaluates the current values when called?

lib/optimizely/odp/odp_config.rb
```diff
@@ -0,0 +1,122 @@
...
+
+        return false if @api_key == api_key && @api_host == api_host && @segments_to_check == segments_to_check
```
zashraf1985, line 49: Returning from within the synchronized block might sometimes cause issues with locks. I am not sure about this though. You can do your research and see what makes sense. I just wanted to point it out.

lib/optimizely/odp/odp_segment_manager.rb
```diff
@@ -0,0 +1,120 @@
...
+module Optimizely
+  class OdpSegmentManager
+    # Schedules connections to ODP for audience segmentation and caches the results
+    attr_reader :odp_config, :segments_cache, :zaius_manager, :logger
+
+    def initialize(cache_size, cache_timeout_in_secs, odp_config, api_manager = nil, logger = nil, proxy_config = nil)
+      @odp_config = odp_config
+      @logger = logger || NoOpLogger.new
+      @zaius_manager = api_manager || ZaiusGraphQLApiManager.new(logger: @logger, proxy_config: proxy_config)
+      @segments_cache = Optimizely::LRUCache.new(cache_size, cache_timeout_in_secs)
```
zashraf1985, line 31: do we want an ability to pass an implementation from outside?

lib/optimizely/odp/odp_segment_manager.rb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    def fetch_qualified_segments(segment_request)
+      odp_api_key = @odp_config&.api_key
```
zashraf1985, line 35: You could probably use `@odp_enabled` here?

lib/optimizely/odp/odp_segment_manager.rb
```diff
@@ -0,0 +1,120 @@
...
+
+      unless odp_api_host && odp_api_key
+        @logger.log(Logger::ERROR, 'api_key/api_host not defined')
+        segment_request.segments = nil
+        return
+      end
+      segments_to_check = @odp_config&.segments_to_check
+
+      unless segments_to_check&.size&.positive?
+        segment_request.segments = []
```
zashraf1985, line 46: A debug log might be helpful here

lib/optimizely/odp/odp_segment_manager.rb
```diff
@@ -0,0 +1,120 @@
...
+          segment_request.segments = segments
+          return
+        end
+      end
+
+      Thread.new do
+        segments = @zaius_manager.fetch_segments(odp_api_key, odp_api_host, segment_request.user_key, segment_request.user_value, segments_to_check)
+        @segments_cache.save(cache_key, segments) unless segments.nil? || ignore_cache
+        segment_request.segments = segments
+      end
```
zashraf1985, line 69: Just curious why do we need a new thread here, the code that follows depends on results from this request so why can't it be a blocking call? The calling thread will be waiting anyway.

zashraf1985 (changes requested):
Looks Great! Suggested a few changes.

lib/optimizely/odp/odp_segment_manager.rb
andrewleap-optimizely: yea good catch

lib/optimizely/odp/odp_segment_manager.rb
andrewleap-optimizely: Yea no you're right, my bad. I've removed this functionality. 👍

jaeopt (changes requested):
Looks good. A few small changes suggested.
Some ruby contexts I cannot cover :)

jaeopt (approved):
LGTM

zashraf1985 (approved):
Looks Great!

PR #312, opened 2022-09-12
feat: add odp event manager

andrewleap-optimizely:
Summary
-------
OdpEventManager provides internal services for sending ODP events to the ODP server. 
- adds default values to identifiers and data fields for all events.
- stores requested events if they cannot be sent to the server.
- stored events are dispatched when resources are available.

Test plan
---------
- added odp_event_manager_spec.rb

Ticket
------
[OASIS-8445](https://optimizely.atlassian.net/browse/OASIS-8445)

### DIFF
--- a/lib/optimizely/helpers/constants.rb
+++ b/lib/optimizely/helpers/constants.rb
@@ -385,7 +385,8 @@ module Optimizely
       ODP_LOGS = {
         FETCH_SEGMENTS_FAILED: 'Audience segments fetch failed (%s).',
         ODP_EVENT_FAILED: 'ODP event send failed (%s).',
-        ODP_NOT_ENABLED: 'ODP is not enabled.'
+        ODP_NOT_ENABLED: 'ODP is not enabled.',
+        ODP_NOT_INTEGRATED: 'ODP is not integrated.'
       }.freeze
 
       DECISION_NOTIFICATION_TYPES = {
@@ -428,6 +429,13 @@ module Optimizely
         UNDETERMINED: 'UNDETERMINED',
         INTEGRATED: 'INTEGRATED',
         NOT_INTEGRATED: 'NOT_INTEGRATED'
+      }.freeze
+
+      ODP_EVENT_MANAGER = {
+        DEFAULT_QUEUE_CAPACITY: 10_000,
+        DEFAULT_BATCH_SIZE: 10,
+        DEFAULT_FLUSH_INTERVAL: 1,
+        DEFAULT_RETRY_COUNT: 3
       }.freeze
 
       HTTP_HEADERS = {
--- a/lib/optimizely/helpers/constants.rb
+++ b/lib/optimizely/helpers/constants.rb
@@ -434,7 +434,7 @@ module Optimizely
       ODP_EVENT_MANAGER = {
         DEFAULT_QUEUE_CAPACITY: 10_000,
         DEFAULT_BATCH_SIZE: 10,
-        DEFAULT_FLUSH_INTERVAL: 1,
+        DEFAULT_FLUSH_INTERVAL_SECONDS: 1,
         DEFAULT_RETRY_COUNT: 3
       }.freeze
 
--- a/lib/optimizely/helpers/constants.rb
+++ b/lib/optimizely/helpers/constants.rb
@@ -424,6 +424,12 @@ module Optimizely
         REQUEST_TIMEOUT: 10
       }.freeze
 
+      ODP_CONFIG_STATE = {
+        UNDETERMINED: 'UNDETERMINED',
+        INTEGRATED: 'INTEGRATED',
+        NOT_INTEGRATED: 'NOT_INTEGRATED'
+      }.freeze
+
       HTTP_HEADERS = {
         'IF_MODIFIED_SINCE' => 'If-Modified-Since',
         'LAST_MODIFIED' => 'Last-Modified'
--- a/lib/optimizely/odp/odp_config.rb
+++ b/lib/optimizely/odp/odp_config.rb
@@ -32,7 +32,7 @@ module Optimizely
       @api_host = api_host
       @segments_to_check = segments_to_check
       @mutex = Mutex.new
-      @odp_state = @api_host.nil? && @api_key.nil? ? ODP_CONFIG_STATE[:UNDETERMINED] : ODP_CONFIG_STATE[:INTEGRATED]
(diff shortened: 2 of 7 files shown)


jaeopt (changes requested):
It looks great! A few small changes suggested.

spec/odp/odp_event_manager_spec.rb
```diff
@@ -0,0 +1,486 @@
...
+      event[:data][:data_source] = 'my-app'
+      odp_event = Optimizely::OdpEvent.new(**event)
+
+      allow(event_manager.zaius_manager).to receive(:send_odp_events).once.with(api_key, api_host, [odp_event]).and_return(false)
+      event_manager.start!
+
+      event_manager.send_event(**event)
+      event_manager.flush
+      sleep(0.1) until event_manager.instance_variable_get('@event_queue').empty?
+
```
jaeopt, line 361: Not validating the data valye is overriden?

jaeopt (changes requested):
Those small changes again :)

spec/odp/odp_event_manager_spec.rb
andrewleap-optimizely: You weren't convinced by the name of the test? 😉

lib/optimizely/odp/odp_config.rb
```diff
@@ -41,14 +44,19 @@ def initialize(api_key = nil, api_host = nil, segments_to_check = [])
...
     def update(api_key = nil, api_host = nil, segments_to_check = [])
+      updated = false
       @mutex.synchronize do
-        break false if @api_key == api_key && @api_host == api_host && @segments_to_check == segments_to_check
-
-        @api_key = api_key
-        @api_host = api_host
-        @segments_to_check = segments_to_check
-        break true
+        @odp_state = api_host.nil? && api_key.nil? ? ODP_CONFIG_STATE[:NOT_INTEGRATED] : ODP_CONFIG_STATE[:INTEGRATED]
```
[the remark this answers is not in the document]
andrewleap-optimizely: Agreed

jaeopt (approved):
LGTM

lib/optimizely/odp/odp_event_manager.rb
```diff
@@ -0,0 +1,249 @@
...
+    def running?
+      @thread && !!@thread.status
+    end
+
+    private
+
+    def run
+      loop do
+        @mutex.synchronize do
+          @received.wait(@mutex, queue_timeout) if @event_queue.empty?
```
zashraf1985, line 146: Just curios if the `sizedQueue` blocks for the event too? If yes, could you use that instead of an extra ConditionVariable to signal? I am not asking to change the implementation. This looks good. Just throwing out a suggestion if it makes sense and you like it, feel free to use it other wise this implementation is good too.

lib/optimizely/odp/odp_event_manager.rb
```diff
@@ -0,0 +1,249 @@
...
+
+      event = Optimizely::OdpEvent.new(type: type, action: action, identifiers: identifiers, data: data)
+      dispatch(event)
+    end
+
+    def stop!
+      return unless running?
+
+      begin
+        @event_queue.push(:SHUTDOWN_SIGNAL, non_block: true)
```
zashraf1985, line 120: The events that are received after this `SHUTDOWN` signal should be dropped. It appears that the event rejection depends on `isRunning` which will not become false unless `SHUTDOWN_SIGNAL` has been processed and the thread is done with. If my understanding is correct, it might be helpful to add a boolean `shutdown_started` or something along those lines which can be immediately set to `true` as soon as `stop` is called and the event manager should start rejecting events as soon as `shutdown_started` becomes true. What do you think?

zashraf1985 (changes requested):
Great Work! Your code is always beautiful and well crafted. This looks great overall. I have just added a few random suggestions here and there.

lib/optimizely/odp/odp_event_manager.rb
andrewleap-optimizely: Yea that's how the python version is implemented; unfortunately ruby's sizedQueue does not have timeout functionality, so we are forced to build our own using the CondVar 👎

lib/optimizely/odp/odp_event_manager.rb
andrewleap-optimizely: SizedQueue has a close method that seems designed for this exact scenario- seems like a good solution?

lib/optimizely/helpers/constants.rb
```diff
@@ -424,6 +425,19 @@ module Constants
...
+      ODP_CONFIG_STATE = {
+        UNDETERMINED: 'UNDETERMINED',
+        INTEGRATED: 'INTEGRATED',
+        NOT_INTEGRATED: 'NOT_INTEGRATED'
+      }.freeze
+
+      ODP_EVENT_MANAGER = {
+        DEFAULT_QUEUE_CAPACITY: 10_000,
+        DEFAULT_BATCH_SIZE: 10,
+        DEFAULT_FLUSH_INTERVAL: 1,
```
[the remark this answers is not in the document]
andrewleap-optimizely: Makes sense to me

lib/optimizely/odp/odp_config.rb
```diff
@@ -30,6 +32,7 @@ def initialize(api_key = nil, api_host = nil, segments_to_check = [])
       @api_host = api_host
       @segments_to_check = segments_to_check
       @mutex = Mutex.new
+      @odp_state = @api_host.nil? || @api_key.nil? ? ODP_CONFIG_STATE[:UNDETERMINED] : ODP_CONFIG_STATE[:INTEGRATED]
```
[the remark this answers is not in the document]
andrewleap-optimizely: I don't have a strong opinion either way, but apparently Ruby does because this is explicitly called out in the style guide:
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/TernaryParentheses 😄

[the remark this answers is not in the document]
zashraf1985: Ahhh!! Rubocop is brutal! I take my suggestion back :)

lib/optimizely/odp/odp_event_manager.rb
zashraf1985: Looks pretty neat to me 👍

zashraf1985 (approved):
LGTM!

jaeopt (approved):
LGTM

PR #314, opened 2022-09-22
feat: add odp manager

andrewleap-optimizely:
## Summary
Provides all internal services related to the ODP integration:
- OdpSegmentManager
- OdpEventManager
- OdpSegmentsCache

Internal API:
- fetch_qualified_segments
- identify_user
- send_event

## Test plan
- odp_manager_spec.rb

## Ticket
- [OASIS-8454](https://jira.sso.episerver.net/browse/FSSDK-8454)

### DIFF
--- a/lib/optimizely/odp/odp_manager.rb
+++ b/lib/optimizely/odp/odp_manager.rb
@@ -28,6 +28,9 @@ require_relative 'odp_event_manager'
 module Optimizely
   class OdpManager
     ODP_LOGS = Helpers::Constants::ODP_LOGS
+    ODP_MANAGER_CONFIG = Helpers::Constants::ODP_MANAGER_CONFIG
+    ODP_CONFIG_STATE = Helpers::Constants::ODP_CONFIG_STATE
+
     def initialize(disable:, segments_cache: nil, segment_manager: nil, event_manager: nil, logger: nil)
       @enabled = !disable
       @segment_manager = segment_manager
@@ -36,7 +39,7 @@ module Optimizely
       @odp_config = OdpConfig.new
 
       unless @enabled
-        @logger.log(Logger::DEBUG, 'ODP is disabled')
+        @logger.log(Logger::INFO, ODP_LOGS[:ODP_NOT_ENABLED])
         return
       end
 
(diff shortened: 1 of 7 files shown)


lib/optimizely/odp/odp_manager.rb
```diff
@@ -0,0 +1,125 @@
...
+      if @odp_config.odp_state == Helpers::Constants::ODP_CONFIG_STATE[:UNDETERMINED]
+        @logger.log(Logger::ERROR, 'Cannot fetch segments before the datafile has loaded.')
+        return nil
+      end
+
+      @segment_manager.fetch_qualified_segments(Helpers::Constants::ODP_MANAGER_CONFIG[:KEY_FOR_USER_ID], user_id, options)
+    end
+
+    def identify_user(user_id:)
+      send_event(
```
jaeopt, line 79: Can we call "@event_manager.send_event()" directly?
"identify_user" is implicit call by SDK while "send_event" explicit call by the client, so errors can be handled differently. When ODP_DISABLED, "identify_user" can drop silently (with debug log) and can skip parameter validation, while "send_event" should log error.

jaeopt (changes requested):
Looks good! A few changes suggested.

lib/optimizely/odp/odp_manager.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    def fetch_qualified_segments(user_id:, options:)
+      # Returns qualified segments for the user from the cache or the ODP server if not in the cache.
+      #
+      # @param user_id - The user id.
+      # @param options - An array of OptimizelySegmentOptions used to ignore and/or reset the cache.
+      #
+      # @return - Array of qualified segments or nil.
+      options ||= []
+      unless @enabled && @segment_manager
```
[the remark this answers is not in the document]
andrewleap-optimizely: True, I'm just used to convincing the type checker... but ruby does not care, so it's removed 🥳

lib/optimizely/odp/odp_manager.rb
andrewleap-optimizely: Makes sense to me

lib/optimizely/odp/odp_manager.rb
```diff
@@ -0,0 +1,144 @@
...
+
+      unless Helpers::Validator.odp_data_types_valid?(data)
+        @logger.log(Logger::ERROR, ODP_LOGS[:ODP_INVALID_DATA])
+        return
+      end
+
+      @event_manager.send_event(type: type, action: action, identifiers: identifiers, data: data)
+    end
+
+    def update_odp_config(api_key, api_host, segments_to_check)
```
zashraf1985, line 124: This is great! I am just curios who will call this function? should we add a notification listener here and make odpmanager self update itself? Or are we expecting to do something like that and trigger updates from the outside by calling this method?

zashraf1985 (approved):
Looks Great! I have just asked one question that i am curios about but that does not block this PR from my side.

lib/optimizely/odp/odp_manager.rb
andrewleap-optimizely: Yea the idea is to register the listener at client initialization and have the callback be a client method, thus avoiding having to pass in the `NotificationListener` and the `ProjectConfig`. Seems to make sense because the callback needs access to both `ProjectConfig` and `OdpConfig`. Thoughts/opinions?

jaeopt (approved):
LGTM

PR #333, opened 2023-04-10
[FSSDK-7825] fix errant test logging

andrewleap-optimizely:
## Summary
Changed `DatafileProjectConfig.create` to use the provided logger in case of error.
- Existing functionality assumed there was no valid logger. This method is always called after logger has been validated.
- If the user has provided a logger, the expectation is that the logger will be used for all logging.

## Test plan
all tests pass

## Ticket
[FSSDK-7825](https://jira.sso.episerver.net/browse/FSSDK-7825)


Mat001 (approved):
Looks good

PR #335, opened 2023-04-21
[FSSDK-9130] fix: invalid identifier error code

andrewleap-optimizely:
## Summary
- Aligned with error code used by ODP

## Test plan
all tests should pass

## Ticket
[FSSDK-9130](https://jira.sso.episerver.net/browse/FSSDK-9130)

### DIFF
--- a/lib/optimizely/odp/odp_segment_api_manager.rb
+++ b/lib/optimizely/odp/odp_segment_api_manager.rb
@@ -79,24 +79,26 @@ module Optimizely
       end
 
       if response.include?('errors')
-        error_class = response['errors']&.first&.dig('extensions', 'classification') || 'decode error'
-        if error_class == 'InvalidIdentifierException'
+        error = response['errors'].first if response['errors'].is_a? Array
+        error_code = extract_component(error, 'extensions', 'code')
+        if error_code == 'INVALID_IDENTIFIER_EXCEPTION'
           log_failure('invalid identifier', Logger::WARN)
         else
+          error_class = extract_component(error, 'extensions', 'classification') || 'decode error'
           log_failure(error_class)
         end
         return nil
       end
 
-      audiences = response.dig('data', 'customer', 'audiences', 'edges')
+      audiences = extract_component(response, 'data', 'customer', 'audiences', 'edges')
       unless audiences
         log_failure('decode error')
         return nil
       end
 
       audiences.filter_map do |edge|
-        name = edge.dig('node', 'name')
-        state = edge.dig('node', 'state')
+        name = extract_component(edge, 'node', 'name')
+        state = extract_component(edge, 'node', 'state')
         unless name && state
           log_failure('decode error')
           return nil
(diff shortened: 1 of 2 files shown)


lib/optimizely/odp/odp_segment_api_manager.rb
```diff
@@ -79,24 +79,26 @@ def fetch_segments(api_key, api_host, user_key, user_value, segments_to_check)
       end
 
       if response.include?('errors')
-        error_class = response['errors']&.first&.dig('extensions', 'classification') || 'decode error'
-        if error_class == 'InvalidIdentifierException'
+        error = response['errors'].first if response['errors'].is_a? Array
+        error_code = extract_component(error, 'extensions', 'code')
+        if error_code == 'INVALID_IDENTIFIER_EXCEPTION'
           log_failure('invalid identifier', Logger::WARN)
```
jaeopt, line 85: Can we use standard error messages (in TDD)?

```
InvalidSegmentIdentifier | WARNING | Audience segments fetch failed (invalid identifier).
FetchSegmentsFailed | ERROR | Audience segments fetch failed (...). 
```

jaeopt (changes requested):
Looks good. Suggest to use standard error messages

lib/optimizely/odp/odp_segment_api_manager.rb
andrewleap-optimizely: This is the format we're using, extracted out for readability:
https://github.com/optimizely/ruby-sdk/blob/a0b44106216aa9bb478dcc0436b74c3dbdcb568a/lib/optimizely/odp/odp_segment_api_manager.rb#L112-L114

I could rename the method to something like `log_segments_failure` to avoid confusion?

Mat001: A broader question:
ODP changing their API/responses is something that is bound to happen again as services get updated. DO we have a strategy to catch such changes on time without breaking customers' use of SDKs? 
@jaeopt

Mat001: How are we informed about ODP changes? We notice FSC failing? That sounds like a terrible safety catch ;)

Mat001 (approved):
looks ok

lib/optimizely/odp/odp_segment_api_manager.rb
jaeopt: > This is the format we're using, extracted out for readability:
> 
> https://github.com/optimizely/ruby-sdk/blob/a0b44106216aa9bb478dcc0436b74c3dbdcb568a/lib/optimizely/odp/odp_segment_api_manager.rb#L112-L114
> 
> I could rename the method to something like `log_segments_failure` to avoid confusion?

Looks good. I missed it's a part of segment failure logs. A good idea to rename it to avoid confusion :)

jaeopt (approved):
LGTM

PR #338, opened 2023-09-25
[FSSDK-9509] Update http_project_config_manager.rb with soft log warning about polling interval < 30s

Mat001:
Add soft warning that polling interval under 30 s is not recommended.

## Summary
- To synchronize with ter SDKs. Peloton had an issue with setting polling interval too low. So we are adding this soft warning to all SDKs including Ruby.
- See Jira ticket below for more.

## Test plan
Unit tests

## Issues
- [FSSDK-9509](https://jira.sso.episerver.net/browse/FSSDK-9509)

### DIFF
--- a/lib/optimizely/config_manager/http_project_config_manager.rb
+++ b/lib/optimizely/config_manager/http_project_config_manager.rb
@@ -268,7 +268,7 @@ module Optimizely
         return
       end
 
-      unless polling_interval.positive? && polling_interval < 30
+      if polling_interval < 30
         @logger.log(
           Logger::WARN,
           'Polling intervals below 30 seconds are not recommended.'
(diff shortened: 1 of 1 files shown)


lib/optimizely/config_manager/http_project_config_manager.rb
```diff
@@ -268,6 +268,13 @@ def polling_interval(polling_interval)
         return
       end
 
+      unless polling_interval.positive? && polling_interval < 30
```
andrewleap-optimizely, line 271: should probably be:
```suggestion
      if polling_interval < 30
```

andrewleap-optimizely (changes requested):

andrewleap-optimizely (approved):
lgtm

PR #301, opened 2022-05-19
feat: include object id/key in invalid object errors

stoneman:
## Summary
- Include object `id`/`key` in errors when objects not found in datafile
- Modifies invalid object `id`/`key` log messages to make them consistent

Include object `id`/`key` in errors when an object is not found makes them available to the user or the custom error handler. 

One example of why this is useful is that the `key` of an experiment could be used within a custom error handler to fetch the details of the experiment. This would indicate whether the experiment has been paused (in which case the error could be ignored) or archived (in which case the code referencing the experiment should be removed from the application).

This change will impact existing custom error handlers which are expecting an error class rather than an error object instance.

## Test plan

Jira ticket:
https://jira.sso.episerver.net/browse/FSSDK-9439

PR #339, opened 2023-09-29
add identifiers

Mat001:
## Summary
- For details please see this original PR: https://github.com/optimizely/ruby-sdk/pull/301 (@stoneman)
- This PR is a copy of PR301 so that we can run FSC and integration tests that were unable to run by external contributor.

#### Details from the original PR:
- Include object `id`/`key` in errors when objects not found in datafile
- Modifies invalid object `id`/`key` log messages to make them consistent

Include object `id`/`key` in errors when an object is not found makes them available to the user or the custom error handler. 

One example of why this is useful is that the `key` of an experiment could be used within a custom error handler to fetch the details of the experiment. This would indicate whether the experiment has been paused (in which case the error could be ignored) or archived (in which case the code referencing the experiment should be removed from the application).

This change will impact existing custom error handlers which are expecting an error class rather than an error object instance.

## Test plan

Jira ticket:
https://jira.sso.episerver.net/browse/FSSDK-9439

### DIFF
--- a/lib/optimizely/config/datafile_project_config.rb
+++ b/lib/optimizely/config/datafile_project_config.rb
@@ -223,8 +223,9 @@ module Optimizely
       experiment = @experiment_key_map[experiment_key]
       return experiment if experiment
 
-      @logger.log Logger::ERROR, "Experiment key '#{experiment_key}' is not in datafile."
-      @error_handler.handle_error InvalidExperimentError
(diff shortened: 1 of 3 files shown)


Mat001:
@jaeopt Can you please approve this? Andy already approved, It's a copy of the original PR to be able to test FSC and integration tests.

andrewleap-optimizely:
This was merged via #301

Issue #243, opened 2020-03-11
Implement Factory for Creating Optimizely::Project instances to Avoid Passing in a Large Number of Positional Arguments


aagrawal2001:
The constructor for `Optimizely::Project` expects 10 positional arguments. This is pretty unwieldy because it's hard for a caller to remember which argument is which. It also requires a bunch of `nil` values to be passed in if only some of the non-default values are desired.

It would be helpful to introduce a factory that takes in keyword arguments as input and constructs the `Project` instance. Since there are probably existing callers, we can't remove the constructor with positional arguments.

jasonkarns:
It can't be removed, but the existing constructor can be "uplifted".

Moving the constructor to use keyward args would solve this as well as avoid needing a factory:

consider an #initialize signature:

```ruby
    def initialize(
      x_datafile = nil,
      x_event_dispatcher = nil,
      x_logger = nil,
      x_error_handler = nil,
      x_skip_json_validation = false,
      x_user_profile_service = nil,
      x_sdk_key = nil,
      x_config_manager = nil,
      x_notification_center = nil,
      x_event_processor = nil,
      datafile: x_datafile,
      event_dispatcher: x_event_dispatcher,
      logger: x_logger,
      error_handler: x_error_handler,
      skip_json_validation: x_skip_json_validation,
      user_profile_service: x_user_profile_service,
      sdk_key: x_sdk_key,
      config_manager: x_config_manager,
      notification_center: x_notification_center,
      event_processor: x_event_processor,
      **opts
    )
```

To clarify what is happening: all the positional-arg variable names would be renamed with some (ideally consistent) prefix or suffix. This allows the existing variable names to be used as the keys in the keywarg form. Secondly, the _default_ value for all the keywargs is set to the value provided by the positional argument. So positional-arg invocations continue to work (their values get assigned automatically to the keywarg vars). And lastly, because all the positional args are _already_ optional, users would be able to invoke the constructor with _only_ the keywarg style. That allows callers to only specify values for the parameters the wish, without the need for a bevy of `nil` placeholders. The rest of the body of the method remains untouched and its default logic can continue. (Though ideally, a future release would move the defaults into the method signature itself and out of the body. As well as eventually removing the old positional-arg style.)

Tamara-Barum:
Created internal ticket for review: FSSDK-9382

andrewleap-optimizely:
This has been addressed. #342

PR #342, opened 2023-12-07
[FSSDK-9382] switch client init args from positional to keyword

andrewleap-optimizely:
## Summary
The Optimizely constructor uses positional args and since all of them are optional, it can become unwieldy.

This PR changes the positional args to keyword.

## Ticket
[FSSDK-9382](https://jira.sso.episerver.net/browse/FSSDK-9382)

## Issues
https://github.com/optimizely/ruby-sdk/issues/243

### DIFF
--- a/spec/project_spec.rb
+++ b/spec/project_spec.rb
@@ -48,7 +48,7 @@ describe 'Optimizely' do
   let(:version) { Optimizely::VERSION }
   let(:impression_log_url) { 'https://logx.optimizely.com/v1/events' }
   let(:conversion_log_url) { 'https://logx.optimizely.com/v1/events' }
-  let(:project_instance) { Optimizely::Project.new(config_body_JSON, nil, spy_logger, error_handler, false, nil, nil, nil, nil, nil, [], {batch_size: 1}) }
+  let(:project_instance) { Optimizely::Project.new(datafile: config_body_JSON, logger: spy_logger, error_handler: error_handler, event_processor_options: {batch_size: 1}) }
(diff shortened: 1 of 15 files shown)


spec/project_spec.rb
```diff
@@ -48,7 +48,7 @@
   let(:version) { Optimizely::VERSION }
   let(:impression_log_url) { 'https://logx.optimizely.com/v1/events' }
   let(:conversion_log_url) { 'https://logx.optimizely.com/v1/events' }
-  let(:project_instance) { Optimizely::Project.new(config_body_JSON, nil, spy_logger, error_handler, false, nil, nil, nil, nil, nil, [], {batch_size: 1}) }
+  let(:project_instance) { Optimizely::Project.new(datafile: config_body_JSON, logger: spy_logger, error_handler: error_handler, event_processor_options: {batch_size: 1}) }
```
mikechu-optimizely, line 51: Do you want to change `error_handler: error_handler,` to `error_handler:,`?

mikechu-optimizely: ...and in other spots too?  Is the linter looking for this in the SDK like the testapp?

mikechu-optimizely (changes requested):
Some thoughts and suggestions for whatever my Ruby review's worth.

spec/project_spec.rb
andrewleap-optimizely: We still support 3.0 and the shorthand syntax was introduced in 3.1

PR #343, opened 2023-12-07
[FSSDK-9781] warn on duplicate experiment key

andrewleap-optimizely:
## Summary
- Ensure newest duplicate experiment key takes precedence (current behavior)
- log a warning when a duplicate experiment key is found

## Test plan
- added test for dupe experiment

## Ticket
[FSSDK-9781](https://jira.sso.episerver.net/browse/FSSDK-9781)

### DIFF
--- a/spec/optimizely_config_spec.rb
+++ b/spec/optimizely_config_spec.rb
@@ -840,9 +840,11 @@ describe Optimizely::OptimizelyConfig do
       ]
     }
 
-    config_body['experiments'].push(new_experiment)
-    config_body['featureFlags'].push(new_feature)
-    project_config = Optimizely::DatafileProjectConfig.new(JSON.dump(config_body), spy_logger, error_handler)
+    config = OptimizelySpec.deep_clone(config_body)
+
+    config['experiments'].push(new_experiment)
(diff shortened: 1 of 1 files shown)


lib/optimizely/optimizely_config.rb
```diff
@@ -91,6 +92,7 @@ def audiences_map
 
     def experiments_map
       experiments_id_map.values.reduce({}) do |experiments_key_map, experiment|
+        @logger.log(Logger::WARN, "Duplicate experiment keys found in datafile: #{experiment['key']}") if experiments_key_map.key? experiment['key']
```
jaeopt, line 95: Looks good. A question is the order in datafile is managed in experiment_id_map. Just for consistency, we're supposed to return the last experiment in the datafile.

jaeopt (changes requested):
Looks good. A clarification.

lib/optimizely/optimizely_config.rb
andrewleap-optimizely: Yea Ruby iterates over hash values in insertion order, so the last instance of an experiment key will take precedence.

jaeopt (approved):
LGTM

PR #353, opened 2024-12-09
[FSSDK-10765] enhancement: Implement UPS request batching for decideForKeys

FarhanAnjum-opti:
Summary
-------
- Currently, decideForKeys loops over the list of flags and calls decide for each, which in turn calls ups. So for decideForKeys, there is two calls per flag to UPS. This PR batches UPS calls when using decideForKeys.

Test plan
---------

- Existing tests are updated to work with the new change. New tests will be added for batching.

Issues
------

-  [FSSDK-10765](https://jira.sso.episerver.net/browse/FSSDK-10765)


Mat001 (approved):
lgtm

PR #399, opened 2026-05-26
[AI-FSSDK] [FSSDK-12670] Block ODP identify event for single identifier

jaeopt:
## Summary

Fixes ODP identify event to only send when multiple valid identifiers exist. Previously, the SDK always sent an identify event when creating a user context, even with just a single `fs_user_id`. This wastes network calls and pollutes ODP identity graphs with single-identifier records that provide no linking value.

## Changes
- Changed `identify_user` method signature across the identify chain to accept an `identifiers` hash instead of a single `user_id` keyword argument
- Added identifier count check in `OdpManager#identify_user` to skip sending when fewer than 2 valid identifiers are present
- Filters out identifiers with nil or empty string values before counting
- Added debug-level logging when the identify event is skipped due to single identifier
- Updated `OptimizelyUserContext` to construct identifiers hash using the `KEY_FOR_USER_ID` constant
- Updated existing tests and added new tests for single-identifier blocking and empty-value filtering

## Jira Ticket
[FSSDK-12670](https://optimizely-ext.atlassian.net/browse/FSSDK-12670)

## Notes
- Reference implementation: Python SDK PR #513

[FSSDK-12670]: https://optimizely-ext.atlassian.net/browse/FSSDK-12670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/optimizely.rb
+++ b/lib/optimizely.rb
@@ -1018,13 +1018,13 @@ module Optimizely
       @odp_manager.send_event(type: type, action: action, identifiers: identifiers, data: data)
     end
 
-    def identify_user(user_id:)
+    def identify_user(identifiers:)
       unless is_valid
         @logger.log(Logger::ERROR, InvalidProjectConfigError.new('identify_user').message)
         return
       end
 
-      @odp_manager.identify_user(user_id: user_id)
+      @odp_manager.identify_user(identifiers: identifiers)
     end
 
     def fetch_qualified_segments(user_id:, options: [])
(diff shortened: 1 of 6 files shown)


esrakartalOpt (approved):
LGTM 👍  It fixes the ODP failures. Decision holdout failures were handled on another PR.

PR #401, opened 2026-06-25
[AI-FSSDK] [FSSDK-12813] Normalize decision event campaign_id, variation_id, and entity_id

jaeopt:
## Summary
Normalize three ID fields on outgoing decision events — `decisions[].campaign_id`, `decisions[].variation_id`, and impression `events[].entity_id` — so the wire payload is byte-equivalent across SDKs for every decision type (experiment, feature test, rollout, holdout). The normalization path is silent and never drops, defers, or fails event dispatch.

Per the updated FSSDK-12813 spec, `campaign_id` and `entity_id` accept any non-empty string (numeric or opaque, e.g. `"default-12345"`, `"layer_abc"`); the fallback to `experiment_id` fires only when the value is empty string, `nil`, or missing. `variation_id` keeps the stricter numeric-string-only contract — empty/whitespace/non-numeric values normalize to `nil`. Non-string types are out of scope.

## Changes
- Added a shared `Helpers::EventIdValidator` module with two predicates: `non_empty_string?` (used for `campaign_id` and `entity_id`, accepts any non-empty string including opaque IDs) and `numeric_string?` (used for `variation_id`, requires decimal digits only). The normalization helpers apply the spec's substitution rules: `campaign_id` falls back to `experiment_id` only when empty/nil; invalid `variation_id` becomes `nil`.
- Routed every impression decision through the new normalization helper in `EventFactory#create_impression_event_visitor` so the rule applies uniformly to experiment, feature test, rollout, and holdout decisions with no per-type branching.
- Guaranteed that the impression `events[].entity_id` reuses the same normalized `campaign_id` byte-for-byte (impression-only; conversion event entity_id is unchanged).
- Added unit coverage for both predicates and the normalizers (including opaque/whitespace passthrough for `campaign_id` and nil-fallback for `variation_id`) and integration coverage on the EventFactory wire payload covering nil, empty, whitespace, non-string, non-numeric-placeholder, and opaque-prefixed (e.g. `default-12345`) inputs, plus a holdout scenario.
- Updated a pre-existing wire-payload assertion in `spec/project_spec.rb` to expect `variation_id: nil` (was `''`) for the rollout-no-match path, per the variation_id normalization rule (FR-011).

## Jira Ticket
[FSSDK-12813](https://optimizely-ext.atlassian.net/browse/FSSDK-12813)

[FSSDK-12813]: https://optimizely-ext.atlassian.net/browse/FSSDK-12813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/optimizely/event/event_factory.rb
+++ b/lib/optimizely/event/event_factory.rb
@@ -23,6 +23,7 @@ require_relative 'entity/snapshot'
 require_relative 'entity/snapshot_event'
 require_relative 'entity/visitor'
 require 'optimizely/helpers/validator'
+require 'optimizely/helpers/event_id_validator'
 module Optimizely
   class EventFactory
     # EventFactory builds LogEvent objects from a given user_event.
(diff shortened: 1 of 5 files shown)


esrakartalOpt (approved):
LGTM 👍  Fix campaign id, variation id, and entity id for holdouts

